PR #1, opened 2019-04-11
Add super basic docker runner.

mozbhearsum:
Pretty straightforward I think? I added a new target in `package.json` to stop it from complaining about not being able to open a browser.

### DIFF
--- a/.neutrinorc.js
+++ b/.neutrinorc.js
@@ -8,8 +8,8 @@ module.exports = {
       '@neutrinojs/react',
       {
         devServer: {
-          host: process.env.HOST,
-          port: process.env.PORT
+          host: process.env.HOST || 'localhost',
+          port: process.env.PORT || 9000
         },
         html: {
           title: 'Balrog Admin',
--- a/.neutrinorc.js
+++ b/.neutrinorc.js
@@ -7,6 +7,10 @@ module.exports = {
     [
       '@neutrinojs/react',
       {
+        devServer: {
+          host: process.env.HOST,
+          port: process.env.PORT
+        },
         html: {
           title: 'Balrog Admin',
           favicon: `${__dirname}/src/images/favicon.png`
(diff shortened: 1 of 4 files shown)


mozbhearsum:
Done! And I moved the host/port setting to the environment.

.neutrinorc.js
```diff
@@ -7,6 +7,10 @@ module.exports = {
     [
       '@neutrinojs/react',
       {
+        devServer: {
+          host: process.env.HOST,
+          port: process.env.PORT
```
helfi92: These will probably require defaults for when running outside docker. You can do:

```jsx
{
  devServer: {
    // ...
  },
  env: {
    HOST: // ...
    port: 9000,
  },
},
```

This will provide environment variable defaults for when they are not defined. https://master.neutrinojs.org/packages/web/#environment-variables

mozbhearsum: Hmm, using the env section didn't work for me -- I'm guessing those get applied after devServer is read? I found another way, though.

helfi92: My mistake. The port should have been capitalized:

```jsx
env: {
  HOST: // ...
  PORT: 9000,
},
```

helfi92: It turns out the environment variables that are defined in `.neutrinorc.js` only get defined after the file has been consumed. In other words, the default envs won't be available for use in `.neutrinorc.js`. In that case, I think we could do:


```diff
diff --git a/.neutrinorc.js b/.neutrinorc.js
index c20659a..d0f9fe7 100644
--- a/.neutrinorc.js
+++ b/.neutrinorc.js
@@ -1,3 +1,5 @@
+const DEFAULT_PORT = 9000;
+
 module.exports = {
   options: {
     root: __dirname,
@@ -11,8 +13,12 @@ module.exports = {
           title: 'Balrog Admin',
           favicon: `${__dirname}/src/images/favicon.png`
         },
+        devServer: {
+          port: process.env.PORT || DEFAULT_PORT
+        },
         env: {
           BALROG_ROOT_URL: 'https://localhost:8010',
+          PORT: DEFAULT_PORT,
         },
       }
     ],
```

PR #2, opened 2019-04-14
Add Auth0 login

helfi92:


src/App.jsx
```diff
@@ -1,15 +1,64 @@
...
+    },
+    user: null,
+  });
+
+  return (
+    <AuthContext.Provider value={authContext}>
+      <ThemeProvider theme={theme}>
+        <Authorize
+          authorize={authorize}
+          popup
```
mozbhearsum: Can we do this in a new tab instead of a new window? It's a little less jarring that way.

src/components/Dashboard/UserMenu.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+const useStyles = makeStyles(theme => ({
+  avatar: {
+    height: theme.spacing(6),
+    width: theme.spacing(6),
+    padding: 0,
+  },
+}));
+
+export default function UserMenu() {
```
mozbhearsum: We added a "Show Access Token" button to our existing dropdown recently, to make things a bit easier for some folks who run some scripts on their laptop. Can we do similar here (or add some other easy way to copy an access token).

src/components/Dashboard/UserMenu.jsx
```diff
@@ -0,0 +1,66 @@
...
+      ) : (
+        <Button
+          onClick={authorize}
+          size="small"
+          variant="contained"
+          color="secondary">
+          Login
+        </Button>
+      )}
+      <Menu
```
mozbhearsum: Small nit: when this pops up, it entirely covers my avatar, rather than appearing below it.

src/App.jsx
```diff
@@ -1,15 +1,64 @@
...
+          audience={process.env.AUTH0_AUDIENCE}
+          redirectUri={process.env.AUTH0_REDIRECT_URI}
+          responseType={process.env.AUTH0_RESPONSE_TYPE}
+          scope={process.env.AUTH0_SCOPE}
+          render={() => {
+            const session = localStorage.getItem(USER_SESSION);
+
+            if (session) {
+              const user = JSON.parse(session);
+              const expires = new Date(user.expiration);
```
mozbhearsum: Is this supposed to handle refresh tokens? If so, I'm not certain it's working. I tried setting `expires` to be slightly in the future, and AFAICT it never tried to refresh.

src/App.jsx
```diff
@@ -1,15 +1,64 @@
...
+          authorize={authorize}
+          popup
+          domain={process.env.AUTH0_DOMAIN}
+          clientID={process.env.AUTH0_CLIENT_ID}
+          audience={process.env.AUTH0_AUDIENCE}
+          redirectUri={process.env.AUTH0_REDIRECT_URI}
+          responseType={process.env.AUTH0_RESPONSE_TYPE}
+          scope={process.env.AUTH0_SCOPE}
+          render={() => {
+            const session = localStorage.getItem(USER_SESSION);
```
mozbhearsum: It looks like this code only runs when I click "Login". Is it possible to have a background login happen, if I've signed in previously? It's rather annoying to need to click "Login" every time I load the site.

src/App.jsx
helfi92: If we do it in a new tab then we won't be able to use `react-auth-components` unfortunately since the refresh token rely on the `Authorize` component so that it can be updated.

helfi92:
> I see that the entire authResult is getting stored in localStorage, but I don't see it in my Developer Toolbar for whatever reason (the only thing in there is loglevel:webpack-dev-server). However, there is a session in my Cookies. Is that expected? I'm guessing there's some magic here that I'm not familiar with...

Should be fixed now.

src/App.jsx
helfi92: > Is this supposed to handle refresh tokens?

It should but I don't think it was working before. I pushed a fix. I'll test it tomorrow.

src/App.jsx
mozbhearsum: Okay, not a blocker.

src/components/Dashboard/UserMenu.jsx
helfi92: Good idea. I filed https://github.com/mozilla-frontend-infra/balrog-ui/issues/4.

src/App.jsx
helfi92: The timer for renewal uses the expiration token in localStorage. This is done in https://github.com/mozilla-frontend-infra/react-auth0-components/blob/99c32611abd9bc5ada5ca5daa8a944d08fa055d3/src/components/Authorize/index.jsx#L264. I don't think setting `expires` here will affect the renewal timer.

PR #8, opened 2019-05-15
Add layout for signoffs view

helfi92:


helfi92:
Screen Shot 2019-05-15 at 4 16 59 PM

mozbhearsum:
I didn't get a chance to dive into the code today, but from looking at and playing around with the UI it seems pretty good. The data is presented in a much more clear way (unlike the current UI). There's a couple of things that aren't totally obvious to me though:
- Is the `<LinkIcon>` in the card headers supposed to link to the edit/add page? I was expecting a permalink to the current state of a particular set of Required Signoffs based on the icon. If it's meant to link to the add/edit page, perhaps one of the pencil icons would be better?
- Is the header link meant to bring us to a page that allows you to edit all of the Required Signoffs? (I realize this part isn't implemented to the point where existing state is shown yet, just want to make sure I understand the intent).

helfi92:
> Is the <LinkIcon> in the card headers supposed to link to the edit/add page

Correct. Clicking on it shows an empty form but ideally the form would be pre-filled.

> perhaps one of the pencil icons would be better?

Yes, that might be a better choice. I'll give it a try.

> Is the header link meant to bring us to a page that allows you to edit all of the Required Signoffs

Which header link? Does my answer to the `<LinkIcon /`> question answer this question?

> I didn't get a chance to dive into the code today

There's no rush. I still need to find the time to rebase on master :)

> it seems pretty good. 

\o/

mozbhearsum:
> Which header link? Does my answer to the <LinkIcon /> question answer this question?

I think so, but just to confirm: the intent is to have the edit page show you all required signoffs for a the porduct + permissions or channel, and let you edit them at the same time? (As opposed to letting you edit them role-by-role.)

src/utils/getRequiredSignoffs.js
```diff
@@ -0,0 +1,35 @@
+import { OBJECT_NAMES } from './constants';
+import RequiredSignoffs from '../views/RequiredSignoffs/ListSignoffs/RequiredSignoffs';
```
mozbhearsum: It looks like the object being imported here is more of a data model than a view, is it normal for those to live in `views` in the React world? I'll totally defer to you on this, I just want to make sure I understand the architecture.

helfi92:
> The edit view doesn't have an obvious to adjust the number of required signoffs for an existing Role (maybe you can hit the trash icon, and then add a new row for that Role, though?). Eg: if Firefox/release currently requires 2 releng signoffs, how do I change that to 1 releng signoff?

Good question. I plan to revamp this view in my next PR. Basically it will just be

Screen Shot 2019-05-17 at 10 55 11 AM

with a garbage can on the right. Sorry for the confusion. I'll have it clearer in the next PR.

> There's no quick way to go from the edit view back to the required signoffs page. Is the intent for the user to use the browser's back button, or should there be something in the for that? I'll defer to you on this, just calling it out.

I think it's fine to rely on the browser to return to the previous page but what we can do is navigate the user to the previous page after the user saves their new changes.

src/utils/getRequiredSignoffs.js
helfi92: Makes sense. I'll move it to be under `views/RequiredSignoffs`.

helfi92: Done.

PR #11, opened 2019-05-21
Revamp the Signoff view

helfi92:


src/views/RequiredSignoffs/ViewSignoff/index.jsx
```diff
@@ -29,106 +39,250 @@ const useStyles = makeStyles(theme => ({
...
+      const roles = getRolesFromRequiredSignoffs(rs.data);
 
-      setProducts(result.data.product);
+      setRoles(roles);
     });
+  }, [product, channel]);
+
+  const renderRole = (role, index) => (
+    <Grid
+      key={role[2].id}
```
helfi92: It would be nice if the endpoint returned roles with a unique property like an `id` or perhaps a combination of properties that makes an entry unique. The `key` property [is important](https://reactjs.org/docs/lists-and-keys.html#keys) when mapping over a list. Right now `role[2].id` is being generated with `Math.random()`.

src/views/RequiredSignoffs/ViewSignoff/index.jsx
```diff
@@ -29,106 +39,250 @@ const useStyles = makeStyles(theme => ({
...
+
+    return role[2].isAdditionalRole
+      ? setAdditionalRoles(additionalRoles.filter(excludeRole))
+      : setRoles(roles.filter(excludeRole));
+  };
+
+  // TODO: Add save logic
+  const handleSignoffSave = () => {};
+  // TODO: Add delete logic
+  const handleSignoffDelete = () => {};
```
helfi92: @mozbhearsum I was thinking the mutations (hooking up the actions) is probably something you can do. This would allow me to work on the general layout of the other views. Let me know what you think. I plan to tackle the rules page next. Just need to find the time :)

src/views/RequiredSignoffs/ViewSignoff/index.jsx
testbhearsum: `role` itself should be unique for any given product/channel or product permissions (it's part of the primary key in the table it's stored in). That should make it OK here, unless I'm missing something?

mozbhearsum: Erm, clearly that was the wrong account I commented with. It was me, though!

helfi92: By `role`, do you mean the combination of all the properties of a role is what's unique?

helfi92:
@mozbhearsum I pushed a couple of commits. Let me know if that's better :)

src/views/RequiredSignoffs/ViewSignoff/index.jsx
mozbhearsum: Yeah, sorry - if we're talking about a response with a list of required signoffs, eg:
```
{"channel":"esr","data_version":1,"product":"SystemAddons","role":"relman","signoffs_required":1},
{"channel":"esr","data_version":1,"product":"SystemAddons","role":"sysaddon","signoffs_required":1},{"channel":"release","data_version":1,"product":"SystemAddons","role":"relman","signoffs_required":1},
```

...product+channel+role is unique.

For permissions, product+role is unique.

src/views/RequiredSignoffs/ViewSignoff/index.jsx
mozbhearsum: That's a great idea. Would you mind doing just one of them first, though? I still don't have a great feel for the structure of React apps, so that would help me bootstrap.

helfi92: Sure thing. I can try to write something up in a separate PR.

src/views/RequiredSignoffs/ViewSignoff/index.jsx
helfi92: Awesome. I will codify that right now :)

helfi92:
> Although, maybe just removing it altogether (and letting the user use the browser controls) might be OK/better?

Agreed :)

src/views/RequiredSignoffs/ListSignoffs/index.jsx
```diff
@@ -58,28 +75,37 @@ function ListSignoffs() {
...
-            </TextField>
+            {permissionChanges && (
+              <Typography gutterBottom variant="h5">
+                Changes to Permissions
+              </Typography>
+            )}
+            {!permissionChanges && rulesOrReleasesChanges && (
+              <Typography gutterBottom variant="h5">
+                Changes to Rules / Releases
+              </Typography>
```
helfi92: This will make sure the title is still aligned horizontally with the dropdown when the permissions section is hidden.

PR #14, opened 2019-05-27
Add mutation sample skeleton

helfi92:


src/views/RequiredSignoffs/ViewSignoff/index.jsx
```diff
@@ -112,9 +119,19 @@ function ViewSignoff({ isNewSignoff, ...props }) {
   };
 
   // TODO: Add save logic
-  const handleSignoffSave = () => {};
+  const handleSignoffSave = async () => {
+    // Call the save endpoint with the appropriate parameters
```
mozbhearsum: So we'll probably want another function in https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/services/requiredSignoffs.js to do the actual work of saving?

helfi92: Yeah, that's correct.

helfi92:
@mozbhearsum Let me know if there is anything I can do to help with hooking up the mutations.

PR #15, opened 2019-05-29
Add rules layout page

helfi92:


helfi92:
So far:

Screen Shot 2019-05-29 at 1 47 27 PM

helfi92:
@mozbhearsum I added a screenshot of what I am leaning towards for the rule card. Let me know what you think and whether you think some things should be changed.

mozbhearsum:
> @mozbhearsum I added a screenshot of what I am leaning towards for the rule card. Let me know what you think and whether you think some things should be changed.

I like it so far! Using a proper diff view scheduled changes is a clever idea. There's some primary still to be added, obviously. Here's a few ideas on what's there now:
* It would be good to just ignore `update_type` in Rules. We actually set it elsewhere now, but have yet to drop the column.
* Within a Rule there's a few different logical groupings of properties, it would be nice to highlight/group them in some way.
  * `buildID`, `buildTarget`, `channel`, `distribution`, `distVersion`, `headerArchitecture`, `locale`, `osVersion`, `product`, `instructionSet`, `memory`, `jaws`, `mig64`, `version` are used to find all the matching rules for a request (eg: https://aus5.mozilla.org/update/6/Firefox/56.0/20170926190823/WINNT_x86-msvc-x64/ach/release-localtest/Windows_NT%2010.0.0.0%20(x64)/ISET:SSE3,MEM:2048,JAWS:0/default/default/update.xml)
  * `priority` is used to further refine the matching rules down to a single rule
  * `backgroundRate`, `fallbackMapping` and `mapping` are used after we filter down to a single rule to determine which release to serve for a request
  * `alias`, `rule_id`, `data_version` and `comment` are informational, only used by automation and humans (not used as part of serving update requests)
* `alias` and `rule_id` serve the same purpose. It would be nice to show `alias` when both exist (and show `rule_id` in a tooltip)

As this is the most important part of the Balrog UI (it gets used daily), I think we'll get other feedback and suggestions when we show it off to RelEng/RelMan/etc.

helfi92:
Screen Shot 2019-06-03 at 11 10 56 AM

Changes:
* Stop displaying `update_type`
* Remove `priority` from the card body. The number is now shown beside the product/channel name. This will make sure a user doesn't see this number when scanning the body of a rule card.
* Group logical grouping properties in separate columns within the card body
* Show alias beside rule ID when applicable

Screen Shot 2019-06-03 at 11 14 02 AM

mozbhearsum:
This looks pretty good, I don't see any reason it can't land. A couple of things that we should fix in follow-ups:
* It takes around a full second to switch to another page, is it possibly to speed that up?
* We should add a show/hide scheduled changes toggle

There will probably be other things as well, once others look at it.

helfi92:
Sounds good. I've filed https://github.com/mozilla-frontend-infra/balrog-ui/issues/23. Thanks!

PR #17, opened 2019-05-31
Part implementation of writing required signoff changes

mozbhearsum:
There's a number of small bugs here still, as well as additional functionality to implement. But I'm at a point now where I'm pretty sure the overall structure will be staying the same, and would appreciate some feedback. Here's a coulpe notable things this change makes:
* Change role data structure to a list of dicts, because there's much more data we need to store to call the right endpoints with the right data.
* Add an `updateRequiredSignoffs` util, because I couldn't make this work without it, for reasons I no longer remember. This is probably OK good to do anyways, because that code got pretty big (although I may be able to refactor a bit).


src/views/RequiredSignoffs/ViewSignoff/index.jsx
```diff
@@ -113,19 +141,25 @@ function ViewSignoff({ isNewSignoff, ...props }) {
   const handleRoleDelete = (role, index) => () => {
     const excludeRole = (entry, i) => !(i === index);
 
-    return role[2].isAdditionalRole
+    removeRole(removedRoles.concat([role]));
```
helfi92: If a user creates 2 additional roles by clicking on the "+" button twice and filling the entries then removing one of them, will `removedRoles` contain one or both roles? I was thinking maybe we could define a utility that finds the removed roles by looking at the `original` and `roles`. What do you think?

src/services/requiredSignoffs.js
```diff
@@ -5,9 +5,26 @@ const getRequiredSignoffs = objectName =>
...
+const updateRequiredSignoff = params => {
+  const { useScheduledChange, scId, ...postData } = params;
+  const type = postData.channel ? 'product' : 'permissions';
+  const scPath = scId ? `/${scId}` : '';
+  const url = useScheduledChange
+    ? `${BASE_URL}/scheduled_changes/required_signoffs/${type}${scPath}`
+    : `${BASE_URL}/required_signoffs/${type}`;
+  // TODO: where can we set this globally as a default?
+  const { accessToken } = JSON.parse(
+    localStorage.getItem('react-auth0-session')
```
helfi92: We can use the `USER_SESSION` constant instead of hard-coding `react-auth0-session`:

```jsx
import { USER_SESSION } from '../utils/constants';
```

src/services/requiredSignoffs.js
```diff
@@ -5,9 +5,26 @@ const getRequiredSignoffs = objectName =>
...
 const getScheduledChanges = objectName =>
   axios.get(`${BASE_URL}/scheduled_changes/${objectName}`);
+const updateRequiredSignoff = params => {
+  const { useScheduledChange, scId, ...postData } = params;
+  const type = postData.channel ? 'product' : 'permissions';
+  const scPath = scId ? `/${scId}` : '';
+  const url = useScheduledChange
+    ? `${BASE_URL}/scheduled_changes/required_signoffs/${type}${scPath}`
+    : `${BASE_URL}/required_signoffs/${type}`;
+  // TODO: where can we set this globally as a default?
```
helfi92: For views, we expose the `withUser` utility that passes the user object as a prop. For utilities, one way to achieve this here is to use the `withUser` in `ViewSignoff/index.jsx ` and pass the user object as an option.

```jsx
const updateRequiredSignoff = (params, options = {}) => {
  const { user } = options;

  // ...
});
```

Another possible way (perhaps even neater) would be to create a `utils/axios.js` that sends the access token whenever the user is logged in and we  are sending a request to our own backend. We would use this new utility as the axios object.

src/views/RequiredSignoffs/utils/updateRequiredSignoffs.js
```diff
@@ -0,0 +1,112 @@
...
+        useScheduledChange = true;
+
+        return ret;
+      }),
+      removedRoles.map(roleItem => {
+        const role = roleItem[0];
+        // eslint-disable-next-line camelcase
+        const signoffs_required = roleItem[1];
+        // eslint-disable-next-line camelcase
+        const data_version = roleItem[2];
```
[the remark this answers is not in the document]
mozbhearsum: Yes, indeed! I actually have that changed locally already. I'll push a new version up once I finish fixing up the other parts.

src/views/RequiredSignoffs/utils/updateRequiredSignoffs.js
```diff
@@ -0,0 +1,112 @@
...
+    product,
+    channel,
+    roles,
+    originalRoles,
+    additionalRoles,
+    removedRoles,
+    isNewSignoff,
+  } = params;
+  let useScheduledChange = !isNewSignoff;
+  const errors = [];
```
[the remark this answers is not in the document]
mozbhearsum: I _think_ we actually can push multiple times if we're updating multiple items at the same time. Eg: changing the number of signoffs for more than one role at once.

src/views/RequiredSignoffs/utils/updateRequiredSignoffs.js
```diff
@@ -0,0 +1,112 @@
...
+          if (
+            value.name === role.name &&
+            value.signoffs_required === newSignoffsRequired
+          ) {
+            skip = true;
+          }
+        });
+
+        if (skip) {
+          return;
```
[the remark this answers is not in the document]
mozbhearsum: Hmmm, good point. It's probably better to return a resolved Promise instead?

src/views/RequiredSignoffs/ViewSignoff/index.jsx
```diff
@@ -113,19 +141,25 @@ function ViewSignoff({ isNewSignoff, ...props }) {
...
-    await sleep(2000);
+    await saveRS({
+      product: productTextValue,
+      channel: channelTextValue,
+      roles,
+      originalRoles,
+      additionalRoles,
+      removedRoles,
+      isNewSignoff,
+    });
```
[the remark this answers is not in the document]
mozbhearsum: It actually does show up, but only briefly. For some reason, `saveAction` has `data` and `error` set to null even after throwing an error. I'm working on debugging this at the moment.

src/views/RequiredSignoffs/utils/updateRequiredSignoffs.js
```diff
@@ -0,0 +1,112 @@
...
+          product,
+          channel,
+          role: role.name,
+          signoffs_required: role.sc
+            ? role.sc.signoffs_required
+            : role.signoffs_required,
+          data_version: role.data_version,
+          useScheduledChange: true,
+          change_type: 'update',
+          when: new Date().getTime() + 5000,
```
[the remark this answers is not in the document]
mozbhearsum: In most cases it will, but if `when` ends up in the past (according to server), it will return a 400. 5s should account for clock skew/request time.

src/views/RequiredSignoffs/ViewSignoff/index.jsx
mozbhearsum: Hmm, I'll look into this.

Are you thinking the utility would be called at the beginning of `updateRequiredSignoffs`?

src/services/requiredSignoffs.js
mozbhearsum: Something would wrap the `axios` object (or maybe each method we use?) and use `transformRequest` to set the header?

src/views/RequiredSignoffs/utils/updateRequiredSignoffs.js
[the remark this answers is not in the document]
mozbhearsum: ...except that I just remembered that `Promise.all` raises the first error it encounters, so even if more are raised we'll only see one? Is that what you were getting at?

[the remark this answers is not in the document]
helfi92: Yeah. `Promise.all` will only show the first error, which is okay in my opinion.

src/views/RequiredSignoffs/utils/updateRequiredSignoffs.js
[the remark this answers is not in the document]
helfi92: You can do `.map().filter(Boolean)`. This would remove empty entries.

mozbhearsum:
This needs a little bit more implementation, and there's a couple of bugs, but I think it's in a landable state now. I'll toss the follow-ups onto the Trello board.

src/views/RequiredSignoffs/ViewSignoff/index.jsx
```diff
@@ -61,46 +62,72 @@ const useStyles = makeStyles(theme => ({
...
-  const handleRoleValueChange = (role, index) => ({ floatValue: value }) => {
-    const setRole = (entry, i) =>
-      i === index ? [entry[0], value, entry[2]] : entry;
+  const handleRoleValueChange = role => ({ floatValue: value }) => {
+    const setRole = entry => {
+      if (entry.name !== role.name) {
+        return entry;
+      }
+
+      const result = Object.assign(entry);
```
helfi92: I think the `no-param-reassign` rule is there to make sure the `arguments` object is not mutated. We could probably just do

```jsx
const setRole = entry => {
  const result = entry;
  
  // rest of logic
};
```

https://stackoverflow.com/a/42399879/6150432

I think there's another place where we can remove `Object.assign`.

src/views/RequiredSignoffs/utils/updateRequiredSignoffs.js
```diff
@@ -0,0 +1,127 @@
...
+
+  if (error) {
+    const config = JSON.parse(error.config.data);
+    const errorMsg =
+      error.response.data.exception ||
+      error.response.data.detail ||
+      'Unknown error';
+
+    throw new Error(`Error updating ${config.role} role: ${errorMsg}`);
+  }
```
helfi92: Very nice!

src/utils/Rules.js
```diff
@@ -1,12 +1,11 @@
...
-const baseUrl = `${process.env.BALROG_ROOT_URL}/api`;
-const getRules = () => axios.get(`${baseUrl}/rules`);
-const getChannels = () => axios.get(`${baseUrl}/rules/columns/channel`);
-const getProducts = () => axios.get(`${baseUrl}/rules/columns/product`);
+const getRules = () => axios.get('/rules');
+const getChannels = () => axios.get('/rules/columns/channel');
+const getProducts = () => axios.get('/rules/columns/product');
 const getHistory = (id, limit, page) =>
-  axios.get(`${baseUrl}/${id}/revisions?${stringify({ limit, page })}`);
+  axios.get(`/${id}/revisions?${stringify({ limit, page })}`);
```
helfi92: I think latest master has some of these rules use `baseUrl`. We would probably need to rebase and fix those functions. Example: https://github.com/mozilla-frontend-infra/balrog-ui/blob/68b3562949c8e947991b7a3671ecb82a78d07f0e/src/utils/Rules.js#L15-L23.

src/utils/axios.js
```diff
@@ -0,0 +1,14 @@
+import * as realAxios from 'axios';
+import { BASE_URL, USER_SESSION } from './constants';
+
+const { accessToken } = JSON.parse(
+  localStorage.getItem(USER_SESSION)
+).authResult;
```
helfi92: I think this might fail when a user is initially logged out. `JSON.parse(null)` throws `TypeError: JSON.parse(...) is null`

src/utils/axios.js
```diff
@@ -0,0 +1,14 @@
...
+const axios = realAxios.create({
+  baseURL: BASE_URL,
+  headers: {
+    Authorization: `Bearer ${accessToken}`,
+  },
+});
+
+export default axios;
```
helfi92: I just thought of something else. We could use [axios interceptors](https://github.com/axios/axios#interceptors) to attach the Authorization header. We would intercept every request and add a Authorization header if the user is logged in. The benefit of this is that we won't need to create a new instance of axios of every request and we would be able to use axios without having to wrap it in a utility.

src/hooks/useAction.js
```diff
@@ -26,9 +26,16 @@ export default action => {
 
       setData(data);
 
-      return { ...state, data };
+      // If we don't explicitly overwrite `error` here, any previous errors
+      // will remain when the caller receives the return value, which
+      // prevents them from accurately detecting success/failure
+      return { ...state, data, error: null };
```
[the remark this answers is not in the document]
mozbhearsum: That's probably a good idea. I didn't see any negative side effects of not doing that (probably due to the redirect), but it's best to be correct.

src/views/RequiredSignoffs/ViewSignoff/index.jsx
mozbhearsum: I'll give this a shot, thank you!

src/utils/axios.js
helfi92: I think we can add it in `src/App.jsx`. `authContext.user` would give us access to the user object.

```jsx
axios.interceptors.request.use(config => {
    const accessToken = // get access token
    config.headers.Authorization =  token;

    return config;
});
```

Issue #13, opened 2019-05-21
Add page name to the app bar


helfi92:
Let's add a page title (view name) somewhere in the appbar. Maybe something like `Balrog Admin | <page-name>`.

![Screen Shot 2019-05-21 at 3 01 46 PM](https://user-images.githubusercontent.com/3766511/58123259-6d193900-7bd9-11e9-8cfd-43417d9af053.png)

PR #21, opened 2019-06-03
Add page name to the app bar

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/13.

### DIFF
--- a/src/views/Releases/Release/index.jsx
+++ b/src/views/Releases/Release/index.jsx
@@ -4,7 +4,7 @@ import Dashboard from '../../../components/Dashboard';
 
 export default function Release() {
   return (
-    <Dashboard>
+    <Dashboard title="Release">
       <Typography variant="subtitle1">Create a Release</Typography>
     </Dashboard>
   );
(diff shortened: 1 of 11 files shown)


src/views/Releases/Release/index.jsx
```diff
@@ -4,7 +4,7 @@ import Dashboard from '../../../components/Dashboard';
 
 export default function Release() {
   return (
-    <Dashboard>
+    <Dashboard title="Release">
```
mozbhearsum: For the specific-object pages like this, what do you think about adding or replacing this with the object name? Eg: `Release: Firefox-67.0-build1` or `Rule: firefox-release`, etc.

helfi92: Makes sense. I'll add that prior to merging.

helfi92: Actually, it's probably not needed since the first field of a view will usually tell the user what view they're looking at. For example, in http://localhost:9000/required-signoffs/Firefox, the first field shows the product as "Firefox". Having that value duplicated is probably unnecessary. I'll merge as is but we can probably consider adding it to the view later if there's confusion. Hope that's ok.

PR #24, opened 2019-06-05
Add switch to include scheduled changes

helfi92:


src/components/RuleCard/index.jsx
```diff
@@ -395,7 +397,12 @@ function RuleCard({ rule, ...props }) {
 }
 
 RuleCard.propTypes = {
-  rule,
+  rule: rule.isRequired,
+  showScheduled: bool,
+};
+
+RuleCard.defaultProps = {
+  showScheduled: false,
```
testbhearsum: `true` would be a better default actually. It's typically have people view the Rules.

helfi92: > It's typically have people view the Rules.

Why would someone want to hide scheduled changes in that case? In the old UI, including scheduled changes would strike text making it a bit hard to read current data. Adding a switch made sense to me. In the new UI, it is added it below the rule so it's not really interfering with the displayed rule. I'm just thinking maybe not having a switch makes more sense for the current design. Thoughts?

mozbhearsum: Oh! Sorry, I failed to realize that the Scheduled Changes are always shown (except the diff). I think you're right that it just makes sense to always show them in that case. Sorry for the confusion.

helfi92: No worries at all :)

PR #27, opened 2019-06-11
Implement Rule deletes

mozbhearsum:
This patch implements rule deletes. I struggled a bit getting the new `deleteRule` util working, as noted in the comments. The current state works, but I'm sure it could be improved.


src/utils/Rules.js
```diff
@@ -1,14 +1,16 @@
 import { stringify } from 'qs';
 import axios from 'axios';
 
+// TODO: should this move to services/Rules.jsx?
```
helfi92: Yeah, I think we can do `services/rules.js` (no caps, no jsx).

src/views/Rules/utils/deleteRule.jsx
```diff
@@ -0,0 +1,22 @@
...
+  // TODO: why can't we index the result to just get the error?
+  // await stops working if we do
+  const result = await tryCatch(
+    deleteRule({
+      ruleId: rule.rule_id,
+      dataVersion: rule.data_version,
+    })
+  );
+
+  if (result[0]) {
```
helfi92: We can make it a bit more readable via

```jsx
const [error, result] = await tryCatch(
  // ...
);

if (error) {

}
```

src/views/Rules/utils/deleteRule.jsx
```diff
@@ -0,0 +1,22 @@
...
+    const errorMsg =
+      result[0].response.data.exception ||
+      result[0].response.data.detail ||
+      'Unknown Error';
```
helfi92: Inserting `console.log('pairs: ', Object.entries(result[0]));` shows:

Screen Shot 2019-06-16 at 4 14 45 PM

which suggests the reason this is failing is because `response` is set to`null`.

Screen Shot 2019-06-16 at 4 15 54 PM

mozbhearsum: I'm guessing that you got this error because the backend wasn't running or available? Either way it should be addressed, just wondering about the context.

src/services/rules.js
```diff
@@ -1,14 +1,16 @@
 import { stringify } from 'qs';
 import axios from 'axios';
 
+// TODO: should this move to services/Rules.jsx?
```
helfi92: We can remove this comment now :)

src/views/Rules/ListRules/index.jsx
```diff
@@ -245,7 +261,11 @@ function ListRules(props) {
                   key={`${rule.product}-${rule.channel}-${rule.rule_id}`}
                   item
                   xs={12}>
-                  <RuleCard key={rule.rule_id} rule={rule} />
+                  <RuleCard
+                    key={rule.rule_id}
+                    rule={rule}
+                    handleRuleDelete={() => handleRuleDelete(rule)}
```
helfi92: `onRuleDelete={handleRuleDelete}`.

src/components/RuleCard/index.jsx
```diff
@@ -388,7 +388,9 @@ function RuleCard({ rule, ...props }) {
       <CardActions className={classes.cardActions}>
         <Button color="secondary">Duplicate</Button>
         <Button color="secondary">Update</Button>
-        <Button color="secondary">Delete</Button>
+        <Button color="secondary" onClick={handleRuleDelete}>
```
helfi92: `<Button color="secondary" onClick={() => onRuleDelete(rule)}>`. Let's also add `onRuleDelete` in the list of props: https://github.com/mozilla-frontend-infra/balrog-ui/blob/5d9cd9a6e3f1eb6fe406d7431c24fb3203641ab1/src/components/RuleCard/index.jsx#L399-L401

Perhaps something like:

```jsx
onRuleDelete: func.isRequired,
```

src/views/Rules/ListRules/index.jsx
```diff
@@ -64,9 +65,15 @@ function ListRules(props) {
   const [scheduledChanges, fetchScheduledChanges] = useAction(
     getScheduledChanges
   );
+  const [deleteAction, delRule] = useAction(deleteRule);
   const isLoading = products.loading || channels.loading || rules.loading;
+  // todo: delete errors aren't being shown
```
helfi92: We should probably create a GitHub issue or something so we can keep track of the action items.

src/views/Rules/utils/deleteRule.jsx
helfi92: Maybe. It works now :)

src/views/Rules/utils/deleteRule.jsx
```diff
@@ -0,0 +1,22 @@
+import tryCatch from '../../../utils/tryCatch';
+import { deleteRule } from '../../../services/rules';
+
+export default async rule => {
+  // TODO: why can't we index the result to just get the error?
+  // await stops working if we do
```
helfi92: This should be possible with:

```jsx
const [error] = await tryCatch(
  deleteRule({
    ruleId: rule.rule_id,
    dataVersion: rule.data_version,
  })
);
```

helfi92:
@mozbhearsum Is there anything blocking this PR from being merged?

PR #33, opened 2019-06-26
Basic layout for user editing UI

mozbhearsum:
This is a basic layout for the user view/edit UI. None of the backend operations are implemented yet, but the frontend mostly works (other than a couple of bugs that should be noted in comment).

It's not super performant, which I think is caused by renderPermissions, so I'd appreciate any ideas about making that better. In particular, typing in fields gets slow, presumably because of all of the rerendering.

Any other ideas to improve layout would be appreciated too, or we can fix it up when we come back to the styling the whole app.


helfi92:
Which branch of the balrog repo should I be using? I used both `master` and your `single-schedule-change-get` but I keep getting `Permission denied to perform the request. ({'code': 'invalid_header', 'description': 'Unable to parse authentication token.'}, 401)` when trying to view http://localhost:9000/users.

mozbhearsum:
> Which branch of the balrog repo should I be using? I used both `master` and your `single-schedule-change-get` but I keep getting `Permission denied to perform the request. ({'code': 'invalid_header', 'description': 'Unable to parse authentication token.'}, 401)` when trying to view http://localhost:9000/users.

That's a strange error. We haven't touched that code in the Balrog repo for awhile.

This patch worked fine for me against the latest Balrog master. Do you have any output from the terminal where Balrog's docker containers were brought up?

helfi92:
`docker-compose logs balrogadmin` shows:

```shell
balrogadmin_1  | 2019-07-03 15:37:34,699 - WARNING - PID: 13 - Request: 140076268472696 - auslib.web.admin.views.base.decorated#69: Permission denied to perform the request. ({'code': 'invalid_header', 'description': 'Unable to parse authentication token.'}, 401)
balrogadmin_1  | [pid: 13|app: 0|req: 66/66] 172.18.0.5 () {52 vars in 1623 bytes} [Wed Jul  3 15:37:34 2019] DELETE /api/rules/741?data_version=1 => generated 243 bytes in 339 msecs (HTTP/1.1 403) 9 headers in 404 bytes (1 switches on core 0)
balrogadmin_1  | [pid: 13|app: 0|req: 67/67] 172.18.0.5 () {42 vars in 1330 bytes} [Wed Jul  3 15:37:57 2019] GET /api/scheduled_changes/rules => generated 3592 bytes in 48 msecs (HTTP/1.1 200) 9 headers in 398 bytes (1 switches on core 0)
balrogadmin_1  | [pid: 13|app: 0|req: 68/68] 172.18.0.5 () {40 vars in 1273 bytes} [Wed Jul  3 15:37:57 2019] HEAD /api/csrf_token => generated 0 bytes in 5 msecs (HTTP/1.1 200) 12 headers in 692 bytes (0 switches on core 0)
balrogadmin_1  | 2019-07-03 15:37:57,725 - WARNING - PID: 13 - Request: 140076259495552 - auslib.web.admin.views.base.decorated#69: Permission denied to perform the request. ({'code': 'invalid_header', 'description': 'Unable to parse authentication token.'}, 401)
balrogadmin_1  | [pid: 13|app: 0|req: 69/69] 172.18.0.5 () {44 vars in 1508 bytes} [Wed Jul  3 15:37:57 2019] POST /api/scheduled_changes/rules/7/enact => generated 243 bytes in 361 msecs (HTTP/1.1 403) 9 headers in 404 bytes (1 switches on core 0)
```

`docker-compose ps` outputs:

```shell
➜  balrog git:(master) ✗ docker-compose ps
        Name                      Command                  State                           Ports
--------------------------------------------------------------------------------------------------------------------
balrog_balrogadmin_1   bash scripts/initdb_and_ru ...   Up (healthy)   7070/tcp
balrog_balrogagent_1   /bin/bash /app/run.sh agent      Up
balrog_balrogdb_1      docker-entrypoint.sh mysqld      Up (healthy)   3306/tcp
balrog_balrogpub_1     /bin/bash /app/scripts/run ...   Up             0.0.0.0:9010->9010/tcp
balrog_balrogui_1      /bin/bash /app/docker-entr ...   Up
balrog_nginx_1         /bin/bash -c envsubst '$NG ...   Up (healthy)   0.0.0.0:8010->8010/tcp,
                                                                       0.0.0.0:8011->8011/tcp
```

_Note:_ I seem to see this error even when doing something like deleting a rule.

helfi92:
Apologies for the delay. I will review soon :)

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,470 @@
...
+                onClick={handleUserDelete}
+              />
+            </SpeedDial>
+          )}
+        </Fragment>
+      )}
+    </Dashboard>
+  );
+}
+
```
helfi92: STR:

1. Navigate to the `balrogadmin` user view

Screen Shot 2019-07-04 at 10 00 00 AM

2. Click the "+" button to add a role
3. Modify the first text field with content "releng" by appending another character (e.g., `z`)

**Expected:** the text field with content "releng becomes "relengz".
**Actual:** the empty text field that was added is now "relengz" and the first text field did not change.

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,470 @@
...
+                  label="Username"
+                  value={username}
+                />
+              </Grid>
+            </Grid>
+            <br />
+            <br />
+            <br />
+            <Typography variant="h5">Roles</Typography>
+            {roles.map(renderRole)}
```
helfi92: Let's make the text field occupy the full width of the available space in the main window:

#### Before

Screen Shot 2019-07-04 at 10 22 09 AM

#### After

Screen Shot 2019-07-04 at 10 17 08 AM

We would need to add the `fullWidth` prop to the `TextField` component, then modify the first grid item to have `xs={11}` and then the second one to `xs={1}`. In order to have the garbage icon to the far right, we would need to add a class name to the second column. I think something like:

```jsx
gridRoleDelete: {
    display: 'flex',
    justifyContent: 'flex-end',
},

// ...

<Grid item xs={11} ... />
<Grid item xs={1} className={classes.gridRoleDelete}  ... />

```

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,470 @@
...
+            <Button
+              className={classes.fullWidth}
+              variant="outlined"
+              onClick={() => handleActionAdd(permission)}>
+              <PlusIcon />
+            </Button>
+          </Grid>
+        )}
+      </Grid>
+    ));
```
helfi92: For the permissions, I think we should be able to use the multi selection feature of our `AutoCompleteText` component.

Example: https://codesandbox.io/s/github/kentcdodds/downshift-examples/tree/master/?module=%2Fsrc%2Fordered-examples%2F04-multi-select.js&moduleview=1
Docs: https://github.com/downshift-js/downshift

_Note: `src/components/AutoComplete` is merely a wrapper around the Downshift package that makes the text field have a material look._

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,470 @@
...
+        setPermissions(permissions);
+        setOriginalPermissions(JSON.parse(JSON.stringify(permissions)));
+      });
+    }
+  }, []);
+
+  const handleUsernameChange = ({ target: { value } }) => setUsername(value);
+  // this currently throws an error about controlled components
+  // (https://reactjs.org/docs/forms.html#controlled-components)
+  // not sure how to fix it yet
```
helfi92: I suspect this is because `role.name` is `undefined`when a user clicks the "+" button. Initially when a user clicks the "+" button, the value prop is set to `undefined` (uncontrolled). Typing in it would make the value prop no longer undefined (controlled) which explains why `A component is changing an uncontrolled input of type text to be controlled` shows up in the console. In order to remove the error, the `TextField` component should be defined (controlled) from the start.

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,470 @@
...
+            />
+          )}
+        </Grid>
+        {row[1] === undefined && <Grid item xs />}
+        {row[1] !== undefined && row[1] !== 'add' && (
+          <Fragment>
+            <Grid item xs>
+              <TextField
+                value={row[1]}
+                style={{ width: '85%' }}
```
helfi92: For readability, I find separating styling from the jsx helps. We could refer to a class to attach styles.

src/components/UserList/index.jsx
```diff
@@ -0,0 +1,28 @@
...
+    display: 'block',
+  },
+}));
+
+export default function UserList(props) {
+  const classes = useStyles();
+  const { users } = props;
+
+  return (
+    <GridList className={classes.gridlist}>
```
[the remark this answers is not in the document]
mozbhearsum: It sounds like you're saying just to kill the Grid entirely here, and have a bunch of `<Card>`s as siblings directly beneath the `<Dashboard>`?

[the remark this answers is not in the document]
helfi92: Actually, I was thinking of removing the usage of the Grid and Card altogether. I think I wrote this comment before https://github.com/mozilla-frontend-infra/balrog-ui/pull/33#discussion_r300405853. Thoughts?

src/components/UserCard/index.jsx
```diff
@@ -0,0 +1,91 @@
...
+                  holds the {getRolesString(Object.keys(roles))}
+                </ListItemText>
+              </ListItem>
+            )}
+          </List>
+        </CardContent>
+      </Card>
+    </GridListTile>
+  );
+}
```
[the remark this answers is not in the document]
mozbhearsum: My idea with this was to make it easier to figure out who has access to what, at a glance. If the only place this information is shown is in the specific user pages, it's hard to answer questions like "Who can help me ship SystemAddons/Thunderbird/etc.?". I'm happy to have an alternative way to address this use case, though.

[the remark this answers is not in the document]
helfi92: Ah I see. I would of recommended adding a toolbar of filters similar to the one in https://codetribute.mozilla.org/projects/taskcluster but it might add too much complexity for such a simple view. Given that the list of users won't be > 100, I guess it should be fine to keep as is. The only change in that case would be to replace GridList with Grid. Thanks for the explanation!

src/components/UserList/index.jsx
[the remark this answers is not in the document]
mozbhearsum: I think I'd like to keep the Cards unless we have another idea for satisifying the use case I talked about in that thread.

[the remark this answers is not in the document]
helfi92: Yep, cards sounds good to me.

src/views/Users/ViewUser/index.jsx
mozbhearsum: Thank you!!

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,470 @@
...
+                tooltipTitle="Delete Signoff"
+                onClick={handleUserDelete}
+              />
+            </SpeedDial>
+          )}
+        </Fragment>
+      )}
+    </Dashboard>
+  );
+}
```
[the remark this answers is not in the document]
mozbhearsum: Looks like the root cause of this was the uncontrolled component issue.

src/views/Users/ViewUser/index.jsx
mozbhearsum: I may be misunderstanding what you're suggesting, but I'm not sure multiselect is a great idea here. Not all permissions support the same product and/or action restrictions. And where they do, we don't always grant them the same to some accounts. For example, one account is production has the "rule" permission with the "modify" action restriction, but also the "release" permission with the "create" and "modify" action restrictions.

helfi92: Is the stuff we're trying to do here different than the multi select text fields in the old UI:

Screen Shot 2019-07-05 at 10 56 00 AM

src/views/Users/ViewUser/index.jsx
mozbhearsum: Fixed this with a nested Grid instead.

src/views/Users/ViewUser/index.jsx
mozbhearsum: Oh, my bad! I thought you were talking about using multiselect for the permission name. Let's do this!

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,394 @@
...
+          onSelectedItemsChange={handleRestrictionChange(
+            permission,
+            'products'
+          )}
+          onValueChange={handleRestrictionTextChange(permission, 'productText')}
+          value={permission.metadata.productText}
+          getSuggestions={getSuggestions(products)}
+          label="Product Restrictions"
+          inputProps={{
+            autoFocus: true,
```
helfi92: The `autoFocus` should only be for the first text field. We add this prop so that when the user first enters the view, the first text field has the keyboard cursor already set allowing a user to start typing right away. 

This comment applies to the other text fields as well in this component.

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,394 @@
...
+                  label="Username"
+                  value={username}
+                />
+              </Grid>
+            </Grid>
+            <br />
+            <br />
+            <br />
+            <Typography variant="h5">Roles</Typography>
+            {roles.map(renderRole)}
```
helfi92: Let's make the spacing between "Roles" and the first row similar to the one in the Required Signoff view.


#### http://localhost:9000/users/balrogadmin

Screen Shot 2019-07-15 at 11 03 21 AM

#### http://localhost:9000/required-signoffs/Firefox

Screen Shot 2019-07-15 at 11 03 28 AM

This comment also applies to the "Permissions" header and the row below it. By adding some space on top of the text field, it will also make subsequent text fields have proper spacing similar to the required signoff view.

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,394 @@
...
+            {additionalRoles.map(renderRole)}
+            <Grid item xs className={classes.addGrid}>
+              <Button
+                onClick={handleRoleAdd}
+                className={classes.fullWidth}
+                variant="outlined">
+                <PlusIcon />
+              </Button>
```
helfi92: Let's make the plus button have similar spacing to the required signoff view as well as have its width be similar to the text fields.

Screen Shot 2019-07-15 at 11 08 03 AM

Screen Shot 2019-07-15 at 11 08 51 AM

This comment also applies to the plus button under "Permissions".

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,394 @@
...
+          }}
+        />
+      </Grid>
+      <Grid item xs={1} className={classes.gridDelete}>
+        <IconButton onClick={() => handlePermissionDelete(permission, index)}>
+          <DeleteIcon />
+        </IconButton>
+      </Grid>
+    </Grid>
+  );
```
helfi92: Clicking on the plus button allows a user to add a permission but product restrictions and action restrictions are disabled for that row. Is this expected?

Screen Shot 2019-07-15 at 11 12 42 AM

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,394 @@
...
+import Typography from '@material-ui/core/Typography';
+import IconButton from '@material-ui/core/IconButton';
+import Fab from '@material-ui/core/Fab';
+import SpeedDialAction from '@material-ui/lab/SpeedDialAction';
+import Tooltip from '@material-ui/core/Tooltip';
+import ContentSaveIcon from 'mdi-react/ContentSaveIcon';
+import Button from '@material-ui/core/Button';
+import DeleteIcon from 'mdi-react/DeleteIcon';
+import PlusIcon from 'mdi-react/PlusIcon';
+import ErrorPanel from '@mozilla-frontend-infra/components/ErrorPanel';
```
helfi92: Let's use our own `ErrorPanel` and give the `fixed` prop.

src/utils/Users.js
```diff
@@ -0,0 +1,133 @@
...
+// eslint-disable-next-line import/prefer-default-export
+export {
+  getPermissionString,
+  getRolesString,
+  permissionRestrictionMappings,
+  allPermissions,
+  supportsProductRestriction,
+  supportsActionRestriction,
+  getSupportedActions,
+};
```
helfi92: The filename `Users` gives the impression it's a class or a component which it isn't. Maybe we could rename to something like `userUtils`.

src/utils/Users.js
```diff
@@ -0,0 +1,133 @@
+const niceJoin = array =>
```
helfi92: It would be nice to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ListFormat once it becomes supported in Firefox :) Works in Chrome for now.

helfi92: It took me a bit to understand what `niceJoin` does. Maybe we could rename the function to something like `formatListToLanguage`.

src/utils/Users.js
```diff
@@ -0,0 +1,133 @@
...
+const getRolesString = roles => {
+  const joined = niceJoin(Array.from(roles));
+  let roleStr = 'role';
+
+  if (roles.length > 1) {
+    roleStr = 'roles';
+  }
+
+  return `${joined} ${roleStr}`;
+};
```
helfi92: Nit (feel free to ignore):

```jsx
const getRolesString = roles =>
  `${niceJoin(Array.from(roles))} role${roles.length > 0 ? 's' : ''}`;
```

src/utils/Users.js
```diff
@@ -0,0 +1,133 @@
...
+  'admin',
+  'emergency_shutoff',
+  'rule',
+  'release',
+  'release_read_only',
+  'release_locale',
+  'required_signoff',
+  'permission',
+  'scheduled_change',
+];
```
helfi92: `allPermissions` and `permissionRestrictionMappings` could be in the constants file.

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,394 @@
...
+                  productText: '',
+                  actionText: '',
+                },
+              };
+            }
+          );
+
+          setUsername(userdata.data.data.username);
+          setRoles(roles);
+          setOriginalRoles(JSON.parse(JSON.stringify(roles)));
```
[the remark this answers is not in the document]
mozbhearsum: The reason we're doing it for originalRoles is to ensure we have a separate instance of the object (so that when someone removes an existing role, it doesn't also change originalRoles). So I don't think we need to do the same for roles.

src/components/UserCard/index.jsx
```diff
@@ -0,0 +1,90 @@
...
+            action={<PencilIcon className={classes.pencilIcon} />}
+          />
+        </CardActionArea>
+      </Link>
+      <CardContent>
+        <List>
+          {Object.entries(permissions).map(([permission, details]) => (
+            <ListItem key={permission}>
+              <ListItemIcon>
+                <AccountSupervisorIcon />
```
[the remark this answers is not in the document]
mozbhearsum: That's a fair point, let's do that.

src/views/Users/ViewUser/index.jsx
mozbhearsum: I think I've covered all of this in my upcoming push...

src/views/Users/ViewUser/index.jsx
mozbhearsum: Yup. They'll become enabled (where appropriate) when you enter a valid permission name. For example, the `rule` permission supports both types of restrictions, but `admin` only supports product restrictions.

src/utils/Users.js
mozbhearsum: Yes, that's exactly what I want :sob:

src/utils/Users.js
mozbhearsum: I'm going to keep it the way it is because I find it more readable.

src/views/Users/ViewUser/index.jsx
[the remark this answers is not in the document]
helfi92: Oh I see. To make it more explicit, we could use the `clone` method from ramda:

```jsx
import { clone } from 'ramda';

// ...

setOriginalRoles(clone(roles));
```

src/views/Users/ViewUser/index.jsx
```diff
@@ -0,0 +1,398 @@
...
+                  autoFocus
+                />
+              </Grid>
+            </Grid>
+            <br />
+            <br />
+            <br />
+            <Typography variant="h5">Roles</Typography>
+            {roles.map(renderRole)}
+            {additionalRoles.map(renderRole)}
```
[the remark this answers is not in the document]
mozbhearsum: I moved this out to #83 so we don't forget.

PR #34, opened 2019-06-27
Add create/edit/duplicate rule view

helfi92:


helfi92:
Filed https://github.com/mozilla-frontend-infra/balrog-ui/issues/35 which is a blocker to finishing https://github.com/mozilla-frontend-infra/balrog-ui/issues/35#issuecomment-506394730.

src/views/Rules/Rule/index.jsx
```diff
@@ -1,11 +1,361 @@
...
+              fullWidth
+              label="When"
+              onError={handleDateTimePickerError}
+              helperText={
+                dateTimePickerError ||
+                (scheduleDate < new Date() ? 'Scheduled for ASAP' : undefined)
+              }
+              onDateTimeChange={handleDateTimeChange}
+              value={scheduleDate}
+            />
```
mozbhearsum: I think we need a way to do direct updates to Rules (apologies if I told you otherwise). Day-to-day it's not very useful, but there are times where we're setting up a number of things on test channels (which don't require sign off), and doing that with Scheduled Changes can be cumbersome.

My suggestions would be to use a Scheduled Change by default, but be able to uncheck a box/hide it/set when to null/something else on product+channel combination that don't require signoff. Eg: we shouldn't be able to disable it for Firefox release, but we should be able to for Firefox release-cdntest.

src/views/Rules/Rule/index.jsx
```diff
@@ -1,11 +1,361 @@
...
+
+export default function Rule({ isNewRule, ...props }) {
+  const classes = useStyles();
+  const [rule, setRule] = useState(
+    props.location.state && props.location.state.rule
+      ? props.location.state.rule
+      : initialRule
+  );
+  // 30 seconds - to make sure the helper text "Scheduled for ASAP" shows up
+  const [scheduleDate, setScheduleDate] = useState(addSeconds(new Date(), -30));
```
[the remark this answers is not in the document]
helfi92: That makes a lot of sense. I'll see what I can do to have a way to fix this but we can probably handle this via some logic in a utility function.

src/views/Rules/Rule/index.jsx
helfi92: Is it fair to assume that a direct update is the same thing as an ASAP scheduled change update?

mozbhearsum: Not always. In particular, if a change is scheduled ASAP but it requires signoff from more than one role, it may sit for minutes, hours, or even days before it goes live (because it can't be enacted until it's past the scheduled time _and_ has the necessary signoffs).

When there are no signoffs it's much closer, but it still may take a short time to be enacted because it must hit the scheduled time, and wait for the agent (which polls every 30s) to enact it. So, if ASAP means +30s from current time, it could take up to a minute for the change to be live.

helfi92: I pushed https://github.com/mozilla-frontend-infra/balrog-ui/pull/34/commits/a742995d2e64405c4deefea7a1d42695d9d8ef62 to basically always show the helper text whenever time is within "today". This removes the need to do the `-30` hack.

When an action is triggered like updating a rule, if the scheduled date is in the past, we just need to set it 5 seconds from now to make sure the time is valid. Note that the date picker won't allow a user to pick a past time if the `disablePast` prop is provided so I think we are safe here.

helfi92: > but there are times where we're setting up a number of things on test channels (which don't require sign off), and doing that with Scheduled Changes can be cumbersome

When a user first enters the create page of a rule (http://localhost:9000/rules/create), the date picker will be set to "ASAP" as a default so it won't need to be touched when setting test channels. Also, ASAP will probably mean 5 seconds or less so I think it will only take up to 30 seconds for the change to be live. Does this remove the concern?

mozbhearsum: Let's go with this for now, with the caveat that we may need to come back to it if it becomes too much of a pain for people (it's tough to predict, but we may as well err on the side of simplicity for now).

mozbhearsum: > When an action is triggered like updating a rule, if the scheduled date is in the past, we just need to set it 5 seconds from now to make sure the time is valid. Note that the date picker won't allow a user to pick a past time if the disablePast prop is provided so I think we are safe here.

This makes a lot of sense to me.

PR #36, opened 2019-06-27
Disable settings menu and don't allow access to pages when user is not logged in

mozbhearsum:


src/components/Dashboard/SettingsMenu.jsx
```diff
@@ -5,6 +5,7 @@ import MenuItem from '@material-ui/core/MenuItem';
 import IconButton from '@material-ui/core/IconButton';
 import SettingsOutlineIcon from 'mdi-react/SettingsOutlineIcon';
 import Link from '../../utils/Link';
+import isLoggedIn from '../../utils/isLoggedIn';
```
helfi92: We can use the `withUser` function to have access to the `user` object instead of `isLoggedIn`.

```jsx
import { withUser } from '../../utils/AuthContext';

function SettingsMenu({ user }) {
  // disabled={!user}
}

export default withUser(SettingsMenu);
```

src/components/Dashboard/SettingsMenu.jsx
```diff
@@ -17,6 +18,9 @@ const useStyles = makeStyles(theme => ({
   settingsIcon: {
     fill: '#fff',
   },
+  settingsIconDisabled: {
+    fill: '#aaa',
+  },
```
helfi92: `fill: theme.palette.grey[500],`

src/components/Dashboard/SettingsMenu.jsx
```diff
@@ -31,12 +35,18 @@ export default function SettingsMenu() {
...
         className={classes.settings}
         aria-haspopup="true"
         aria-controls="user-menu"
         aria-label="user menu"
         onClick={handleMenuOpen}>
-        <SettingsOutlineIcon size={24} className={classes.settingsIcon} />
+        <SettingsOutlineIcon
+          size={24}
+          className={
+            isLoggedIn() ? classes.settingsIcon : classes.settingsIconDisabled
```
helfi92:
```jsx
className={user ? classes.settingsIcon : classes.settingsIconDisabled}
```

src/views/RequiredSignoffs/index.jsx
```diff
@@ -1,18 +1,31 @@
...
+            <RouteWithProps
+              key={routeProps.path || 'not-found'}
+              {...routeProps}
+            />
+          ))}
+        </Switch>
+      ) : (
+        <Redirect to="/" />
+      )}
+    </Fragment>
```
helfi92: A view shouldn't be concerned with the redirect flow in my opinion. Instead of having to add the redirect logic in each protected view, what do you think about the following:

```diff
--- a/src/components/RouteWithProps/index.jsx
+++ b/src/components/RouteWithProps/index.jsx
@@ -1,5 +1,6 @@
 import React, { Suspense } from 'react';
-import { Route } from 'react-router-dom';
+import { Route, Redirect } from 'react-router-dom';
+import { withUser } from '../../utils/AuthContext';

 /**
  * Conditionally render a component based on location, with non-react-router
@@ -7,6 +8,8 @@ import { Route } from 'react-router-dom';
  */
 function RouteWithProps(props) {
   const {
+    user,
+    requiresAuth,
     component: Component,
     path,
     exact,
@@ -25,11 +28,15 @@ function RouteWithProps(props) {
       sensitive={sensitive}
       render={({ staticContext, ...renderProps }) => (
         <Suspense fallback={null}>
-          <Component {...renderProps} {...rest} />
+          {requiresAuth && !user ? (
+            <Redirect to="/" />
+          ) : (
+            <Component {...renderProps} {...rest} />
+          )}
         </Suspense>
       )}
     />
   );
 }

-export default RouteWithProps;
+export default withUser(RouteWithProps);
```

Then we could protect individual routes simply by adding `requiresAuth: true` in `route.js`.

_Example: Protecting /users, /roles, /required-signoffs_

```diff
--- a/src/routes.js
+++ b/src/routes.js
@@ -46,15 +46,18 @@ export default [
     component: Users,
     path: '/users',
     routes: userRoutes('/users'),
+    requiresAuth: true,
   },
   {
     component: Roles,
     path: '/roles',
+    requiresAuth: true,
   },
   {
     component: RequiredSignoffs,
     path: '/required-signoffs',
     routes: requiredSignoffsRoutes('/required-signoffs'),
+    requiresAuth: true,
   },
```

mozbhearsum: Makes sense to me, let's do it :)

Issue #20, opened 2019-06-03
Add confirmation dialog for each delete button


helfi92:
Clicking on a "Delete" button or the garbage icon should prompt a confirmation dialog. The content of the dialog could be as simple as:

```
Delete?

Are you sure you want to delete {...}


                                CANCEL  DELETE
```

helfi92:
Re-opening for https://github.com/mozilla-frontend-infra/balrog-ui/pull/93#discussion_r306333025.

divyanshu-rawat:
hey, are there still some buttons that need a confirmation dialog? If yes I can work on it once I have time. what do you say? cc @helfi92

helfi92:
@divyanshu-rawat I _think_ https://github.com/mozilla-frontend-infra/balrog-ui/pull/93#discussion_r306333025 is the only place where we are missing a confirmation dialog. Thank you for offering your help. It is greatly appreciated :)

PR #44, opened 2019-06-28
Dialog for destructive actions

helfi92:
Screen Shot 2019-06-28 at 11 00 30 AM

This will require a rebase once https://github.com/mozilla-frontend-infra/balrog-ui/pull/27 is merged.

Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/20.

### DIFF
--- a/src/views/Rules/ListRules/index.jsx
+++ b/src/views/Rules/ListRules/index.jsx
@@ -12,6 +12,7 @@ import PlusIcon from 'mdi-react/PlusIcon';
 import Dashboard from '../../../components/Dashboard';
 import ErrorPanel from '../../../components/ErrorPanel';
 import RuleCard from '../../../components/RuleCard';
+import DialogAction from '../../../components/DialogAction';
 import Link from '../../../utils/Link';
 import useAction from '../../../hooks/useAction';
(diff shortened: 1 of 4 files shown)


src/views/Rules/ListRules/index.jsx
```diff
@@ -215,6 +222,37 @@ function ListRules(props) {
       rowsPerPage={RULES_ROWS_PER_PAGE}
     />
   );
+  const handleRuleDelete = rule => {
+    setDialogState({
+      ...dialogState,
+      open: true,
+      title: 'Delete Rule?',
+      body: `This will schedule a delete for rule ${rule.rule_id}.`,
```
mozbhearsum: This text isn't necessarily true. Currently, we're only directly deleting rules (no scheduled changes involved). I think eventually we'll delete directly when signoff isn't required, and delete via a scheduled change when it is.

I'm going to update this page to do that in then next day or two.

helfi92:
Rebased.

src/views/Rules/ListRules/index.jsx
helfi92: Sounds good. I'll go ahead and remove "schedule" from the sentence.

PR #47, opened 2019-07-03
Use Scheduled Rule deletes when a delete requires signoff (fixes #45)

### LINKED ISSUE #45: use scheduled changes to delete rules when signoff is required

mozbhearsum:
This enhances the Rules page to use Scheduled deletes when appropriate. I had to do a few extra things to make this possible:
* Pull in the new DateTimePicker & Rule Scheduled Change getter from #34 
* Add support for creating new Rule Scheduled Changes
* Add a helper to determine whether or not a Required Signoff applies to a particular Rule (I'm not sure if I put this in the right place, happy to move it).
* Change the Required Signoffs service to make it consistent with Rules.

A couple of things that I'm not sure about:
* What is the right way to add the Scheduled Change to the existing list of Rules. Should I just Object.assign()?
* Is it OK to be calling `getScheduledChange` from the `handleDialogSubmit` handler? If not, where does it belong?


src/views/Rules/ListRules/index.jsx
```diff
@@ -249,9 +297,35 @@ function ListRules(props) {
...
+
+            i.scheduledChange = sc;
+
+            return i;
+          })
+        );
+      } catch (e) {
+        throw new Error(
+          'Scheduled change created but page failed to properly update page. Please refresh.'
+        );
```
helfi92: If the schedule change was created, is there a reason we would receive an error?

src/views/Rules/ListRules/index.jsx
```diff
@@ -249,9 +297,35 @@ function ListRules(props) {
...
+      }
+    } else {
+      // The rule was directly deleted, just remove it.
+      setRulesWithScheduledChanges(
+        rulesWithScheduledChanges.filter(i => i.rule_id !== dialogRule.rule_id)
```
helfi92: What does `i` stand for? If it's a rule, then a nit would be to probably use `rule` or `r` instead for readability. When I saw `i` I thought it meant index :)

src/views/Rules/ListRules/index.jsx
```diff
@@ -249,9 +297,35 @@ function ListRules(props) {
...
+        const sc = (await getScheduledChange(dialogRule.rule_id)).data
+          .scheduled_changes[0];
+
+        setRulesWithScheduledChanges(
+          rulesWithScheduledChanges.map(i => {
+            if (i.rule_id !== sc.rule_id) {
+              return i;
+            }
+
+            i.scheduledChange = sc;
```
helfi92: I think this will cause a `no-param-reassign` linting error. Alternatives:

1. Use `Object.assign`
2. Create a copy of the rule first then change it

```jsx
rulesWithScheduledChanges.map(r => {
    const rule = { ...r };
});
```

I think the second option is safer since we won't be modifying the original content of the array.

src/utils/requiredSignoffs.js
```diff
@@ -0,0 +1,21 @@
+export default (rule, rs) => {
```
helfi92: Thoughts on renaming this file to `ruleMatchesRequiredSignoff`?

src/views/Rules/ListRules/index.jsx
```diff
@@ -249,9 +297,35 @@ function ListRules(props) {
...
-      rulesWithScheduledChanges.filter(i => i.rule_id !== dialogRule.rule_id)
-    );
+    if (Object.keys(dialogRule.requiredSignoffs).length > 0) {
+      // A change was scheduled, we need to update the card
+      // to reflect that.
+      try {
+        const sc = (await getScheduledChange(dialogRule.rule_id)).data
+          .scheduled_changes[0];
+
+        setRulesWithScheduledChanges(
```
helfi92: `setRulesWithScheduledChanges` should probably happen in the `onComplete` handler. See the review comment above for the explanation.

src/views/Rules/ListRules/index.jsx
```diff
@@ -249,9 +297,35 @@ function ListRules(props) {
       throw error;
     }
 
-    setRulesWithScheduledChanges(
-      rulesWithScheduledChanges.filter(i => i.rule_id !== dialogRule.rule_id)
-    );
+    if (Object.keys(dialogRule.requiredSignoffs).length > 0) {
+      // A change was scheduled, we need to update the card
+      // to reflect that.
+      try {
```
helfi92: No need to wrap in try catch if we make use of the `onError` and `onComplete` props of the dialog.

src/views/Rules/ListRules/index.jsx
mozbhearsum: It's quite unlikely, but if, eg: Balrog went down, or the GET method that we call broke after a deploy, we could get an error.

src/views/Rules/ListRules/index.jsx
mozbhearsum: Item? :smile: I agree that `r` is better.

src/views/Rules/ListRules/index.jsx
mozbhearsum: :+1: It looks like `const rule = { ...r }` will create a shallow copy, but I think that will be OK?

src/utils/requiredSignoffs.js
mozbhearsum: I was thinking of this file as a place to keep general helpers around processing/evaluating required signoffs, which is why I named it this way. Maybe I should export a specific named function rather than this one as the default? What do you think?

helfi92: Exporting a specific named function makes a lot of sense.

src/views/Rules/ListRules/index.jsx
helfi92: You are correct but yeah, I think that's okay for now.

src/views/Rules/ListRules/index.jsx
helfi92: If it breaks or Balrog went down, then there are no guarantees that the schedule change was created right?

mozbhearsum: I think if we get to this point it has been created (because `delRule` would've thrown otherwise). In any case, I removed this message locally, and just let the onError handler show the original exception.

src/views/Rules/ListRules/index.jsx
```diff
@@ -222,23 +243,73 @@ function ListRules(props) {
...
       item: rule,
     });
   };
 
-  const handleDialogError = error => {
-    setDialogState({ ...dialogState, error });
-  };
-
-  const handleDialogClose = () => {
+  const handleDialogClose = result => {
```
helfi92: `onClose` is triggered when the user clicks cancel or the ESC key. `handleDialogClose` will be called since we have`onClose={handleDialogClose}`. This might cause an issue since no result will be passed as a param. We probably want to have a `handleDialogComplete` handler that does the logic then calls `handleDialogClose` at the end. Somthing like:


```jsx
const handleDialogComplete = result => {
  // handle logic here

  this.handleDialogClose();
};

// ...
onComplete={handleDialogComplete}
onClose={handleDialogClose}
```
`handleDialogClose` also gets called when the use

Issue #23, opened 2019-06-04
Rules layout UI enhances


helfi92:
* [x] Virtualize list and remove pagination
* [x] For a rule with scheduled changes, add a dot beside the property names or something that are changing to let the user know that a change to this property is coming soon.
* [x] Add ability to navigate to a card in the rules page via a hash

PR #49, opened 2019-07-05
Rules layout UI enhances

helfi92:
* Remove pagination & virtualize rules view
* Add a red dot beside properties in a card where there are scheduled
changes

Relates to https://github.com/mozilla-frontend-infra/balrog-ui/issues/23.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,8 @@
     "react-hot-loader": "^4",
     "react-number-format": "^4.0.8",
     "react-router-dom": "^5.0.0",
+    "react-virtualized-auto-sizer": "^1.0.2",
+    "react-window": "^1.8.4",
     "unidiff": "^1.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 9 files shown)


helfi92:
> Changing the product/channel filter in the dropdown has just enough of a delay to be annoying, maybe ~500ms.

I'll create an issue to look into this but I think that may be a performance issue with material-ui.

> My overall thought here is that it's a bit weird to have the page-within-a-page when the Rules are the only content. I'm guessing this the virtualization/memoizing necessitates this, however. If so, what do you think about changing the style a bit to make it more natural looking? In particular, I'm thinking that the scrollbar belongs on the far right, and the content should extend all the way to the bottom of the visible window. The cards are a fine size (eg: they don't need to be wider), but it's convenient to be able to scroll without covering any of the content, and it seems unnecessary to not use the last little bit of vertical space.

💯 I forgot about the scrolling. On mac, the scroller is hidden by default unless I scroll so I missed this issue. I'll see what I can do. Thanks for bringing this up!

helfi92:
Pushed a commit to fix the issue. Solution inspired from https://github.com/bvaughn/react-window/issues/30#issuecomment-428868071.

PR #57, opened 2019-07-08
Deep link a rule card

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/23.

### DIFF
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -32,6 +32,24 @@ import getDiff from '../../utils/diff';
 import getDiffedProperties from '../../utils/getDiffedProperties';
 
 const useStyles = makeStyles(theme => ({
+  root: {
+    '& h2': {
+      '& .anchor-link-style': {
+        textDecoration: 'none',
+        opacity: 0,
+        // To prevent the link to get the focus.
+        display: 'none',
+      },
+      '&:hover .anchor-link-style': {
+        display: 'inline-block',
+        opacity: 1,
+        color: theme.palette.text.hint,
+        '&:hover': {
+          color: theme.palette.text.secondary,
+        },
+      },
+    },
+  },
   cardHeader: {
     paddingBottom: 0,
   },
@@ -161,7 +179,7 @@ function RuleCard({ rule, onRuleDelete, ...props }) {
   }, [rule]);
 
   return (
-    <Card spacing={4} {...props}>
+    <Card classes={{ root: classes.root }} spacing={4} {...props}>
       {rule.product && (
         <CardHeader
           classes={{ avatar: classes.cardHeaderAvatar }}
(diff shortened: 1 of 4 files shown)


src/views/Rules/ListRules/index.jsx
```diff
@@ -441,13 +443,41 @@ function ListRules(props) {
...
+      if (ruleId) {
+        const itemNumber = filteredRulesWithScheduledChanges
+          .map(rule => rule.rule_id)
+          .indexOf(ruleId);
+
+        listRef.current.scrollToItem(itemNumber, 'start');
+        setRuleIdHash(hash);
+      }
+    }
+  }, [hash, filteredRulesCount]);
```
mozbhearsum: If I understand effects correctly, depending on `filteredRulesCount` will mean that this code runs anytime it changes (ie: whenever the drop down selection changes). If that's correct, what's the purpose of that?

src/components/RuleCard/index.jsx
```diff
@@ -189,12 +207,18 @@ function RuleCard({ rule, onRuleDelete, ...props }) {
...
-            variant: 'h6',
-          }}
           title={
-            rule.channel ? `${rule.product} : ${rule.channel}` : rule.product
+            <Typography component="h2" variant="h6">
+              {rule.channel
+                ? `${rule.product} : ${rule.channel}`
+                : rule.product}{' '}
+              <a
+                href={`#${rule.rule_id}`}
```
mozbhearsum: Similar to what I discovered in #73, what do we do for pending Rules, which have no rule id yet?

helfi92: I think we should be able to take care of that in https://github.com/mozilla-frontend-infra/balrog-ui/issues/71 where we would be linking to `sc_id`.

src/views/Rules/ListRules/index.jsx
helfi92: At first, `hash` will be defined but `filteredRulesCount` won't be there because the result depends on an asynchronous operation defined in another `useEffect`. The code will go inside this handler but won't go through the [if statement](https://github.com/mozilla-frontend-infra/balrog-ui/pull/57/files/d405b5f66ab1e59453cf6f9c0a50e1af966a0e5f#diff-6bdfeacc9dead52092c38bbdc3958037R459).

Next, when `filteredRulesCount` has been computed, the `useEffect` handler will run again but this time it will go pass the if statement since both `hash` and `filteredRulesCount` are defined.

When a drop down selection changes, the `useEffect` will trigger but it won't satisfy the if statement. The hash is removed when the drop down selection changes.

Hope that helps.

mozbhearsum: Aaaah, I see. Thanks for the explanation.

PR #69, opened 2019-07-08
Centralize error handling for Balrog responses.

mozbhearsum:
This came out of #33, but since it's going to take awhile to get that PR finished up, I'd like to get this merged separately, as it aids any other patches that are doing puts/posts/deletes to Balrog.


src/App.jsx
```diff
@@ -29,6 +29,21 @@ axios.interceptors.request.use(config => {
...
+
+    // If we found a more detailed error message
+    // raise an Error with that instead.
+    if (errorMsg !== null) {
+      throw new Error(errorMsg);
+    }
+
+    throw error;
+  }
+);
```
helfi92: 😍

src/views/RequiredSignoffs/utils/updateRequiredSignoffs.js
```diff
@@ -27,104 +26,93 @@ export default async params => {
...
-          let skip = false;
 
-          originalRoles.forEach(value => {
-            const newSignoffsRequired = role.sc
-              ? role.sc.signoffs_required
-              : role.signoffs_required;
-            const originalSignoffsRequired = value.sc
-              ? value.sc.signoffs_required
-              : value.signoffs_required;
+  await Promise.all(
```
helfi92: Nit: Now that we're not checking for the error in this function, this function doesn't need to have the `async` signature nor `await`. `await` can be replaced with `return`.

Issue #71, opened 2019-07-09
improve display of pending rule inserts


mozbhearsum:
They currently show up only with a diff section:
![image](https://user-images.githubusercontent.com/49649/60901633-48862880-a23c-11e9-9c4d-e01a8693c7a9.png)

Maybe we can show them in a card, and have an indicator that it is still scheduled?

helfi92:
That's a good idea. Exposing the scheduled ID might also be a good idea so we can deep link for example.

mozbhearsum:
It would be good to give them an anchor, too, so we can link back to them when making changes to them.

helfi92:
Agreed. I started working on this a bit. Just need to find the time to finish this.

PR #73, opened 2019-07-09
Implement most of the rule mutations

mozbhearsum:
With this patch we can schedule new rule creations and edits, as well as modify a pending edit.

I still need to figure out how to allow scheduled new rules to be edited though - they don't yet have a rule id, so the /rules/xxx view doesn't work for them. I'm wondering if maybe we need a different route for them that uses sc_id instead of rule_id?


src/views/Rules/Rule/index.jsx
```diff
@@ -117,10 +132,109 @@ export default function Rule({ isNewRule, ...props }) {
...
+  };
+
+  const handleUpdateRule = async () => {
+    const now = new Date();
+    const when =
+      scheduleDate >= now ? scheduleDate.getTime() : now.getTime() + 5000;
+    const data = {
+      alias: rule.alias,
+      backgroundRate: rule.backgroundRate,
+      buildID: rule.buildID,
```
helfi92: Should this be `build_ID`?

src/views/Rules/Rule/index.jsx
```diff
@@ -117,10 +132,109 @@ export default function Rule({ isNewRule, ...props }) {
...
+      buildTarget: rule.buildTarget,
+      channel: rule.channel,
+      comment: rule.comment,
+      distVersion: rule.distVersion,
+      distribution: rule.distribution,
+      fallbackMapping: rule.fallbackMapping,
+      headerArchitecture: rule.headerArchitecture,
+      instructionSet: rule.instructionSet,
+      jaws: rule.jaws,
```
helfi92: I think `jaws` should be a boolean. The text field value is a string. Perhaps `jaws: rule.jaws === 'true'`.

src/views/Rules/Rule/index.jsx
```diff
@@ -117,10 +132,109 @@ export default function Rule({ isNewRule, ...props }) {
...
+      locale: rule.locale,
+      mapping: rule.mapping,
+      memory: rule.memory,
+      mig64: rule.mi64,
```
helfi92: `mig64: rule.mi64 === 'true',`

src/views/Rules/Rule/index.jsx
```diff
@@ -425,10 +549,10 @@ export default function Rule({ isNewRule, ...props }) {
...
           </Tooltip>
-          {scheduleChange.data && scheduleChange.data.data.count > 0 && (
+          {hasScheduledChange && (
             <SpeedDial ariaLabel="Secondary Actions">
               <SpeedDialAction
-                disabled={scheduleChange.loading}
+                disabled={scheduledChangeAction.loading}
                 icon={<DeleteIcon />}
                 tooltipOpen
                 tooltipTitle="Cancel Pending Change"
```
helfi92: STR:

1. Navigate to http://localhost:9000/rules/create
2. Click the save icon button

Expected: An error complaining about one of the required fields not being there
Actual: The error panel shows `update_type: '' is not one of ['major', 'minor', None]`. 2 issues:
1. Update Type should have the `required` prop
2. Update Type should be a menu with inputs "-", 'Minor', 'Major'.

I understand this issue is out of scope from the mutation actions so feel free to do this in a follow-up or we could simply create an issue for it to make sure it is tracked :)

src/views/Rules/Rule/index.jsx
```diff
@@ -83,15 +86,27 @@ export default function Rule({ isNewRule, ...props }) {
...
+    fetchRuleAction.loading ||
+    scheduledChangeAction.loading ||
+    products.loading ||
+    channels.loading;
+  const error =
+    fetchRuleAction.error ||
+    scheduledChangeAction.error ||
+    addSCAction.error ||
+    updateSCAction.error ||
+    deleteSCAction.error;
```
helfi92: Whenever we are doing a mutation, we should be disabling the action buttons to let the user know an action has been triggered as well as stop the user from clicking the same button twice. We can track that in an `actionLoading` variable.

```jsx
const actionLoading = addSCAction.loading || updateSCAction.loading || deleteSCAction.loading;

// ...

{!isLoading && !actionLoading && (
    // action buttons
)}
```

src/views/Rules/Rule/index.jsx
mozbhearsum: I don't think so. I just took a look and the only place (in both the frontend and backend) that uses the underscore is https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/utils/constants.js#L16. I'll update that, instead.

src/views/Rules/Rule/index.jsx
mozbhearsum: Ah, I didn't realize this would be a string. I'll fix this up.

mozbhearsum: We have to support passing null here, so this was slightly more complicated than above, but it's fixed.

mozbhearsum:
This should be ready to go now. I'll need to come back around and fix some of the redirects after #71 is fixed, though.

src/views/Rules/Rule/index.jsx
```diff
@@ -519,11 +519,14 @@ export default function Rule({ isNewRule, ...props }) {
...
+                select
                 label="Update Type"
-                value={defaultToEmptyString(rule.update_type)}
+                value={rule.update_type || 'minor'}
                 name="update_type"
-                onChange={handleInputChange}
-              />
+                onChange={handleInputChange}>
+                <MenuItem value="minor">minor</MenuItem>
+                <MenuItem value="major">major</MenuItem>
```
helfi92: We could capitalize dropdown values:

```jsx
<MenuItem value="minor">Minor</MenuItem>
<MenuItem value="major">Major</MenuItem>
```

src/views/Rules/Rule/index.jsx
```diff
@@ -160,11 +160,11 @@ export default function Rule({ isNewRule, ...props }) {
       fallbackMapping: rule.fallbackMapping,
       headerArchitecture: rule.headerArchitecture,
       instructionSet: rule.instructionSet,
-      jaws: rule.jaws,
+      jaws: rule.jaws === '-' ? null : rule.jaws === 'true',
```
helfi92: Instead of hard-coding `-`, we can use the already defined constant `EMPTY_MENU_ITEM_CHAR`. Same comment applies for the others.

src/views/Rules/Rule/index.jsx
```diff
@@ -392,11 +520,14 @@ export default function Rule({ isNewRule, ...props }) {
             <Grid item xs={12} sm={6}>
               <TextField
                 fullWidth
+                select
```
helfi92: I think we still need to add the `required` prop:

```jsx
<TextField
    fullWidth
    select
    required
    // ...
```

src/views/Rules/Rule/index.jsx
```diff
@@ -412,7 +543,7 @@ export default function Rule({ isNewRule, ...props }) {
           </Grid>
         </Fragment>
       )}
-      {!isLoading && (
+      {!isLoading && !actionLoading && (
```
helfi92: Ah, my bad. This should actually be:

```diff
diff --git a/src/views/Rules/Rule/index.jsx b/src/views/Rules/Rule/index.jsx
index 992d119..042f084 100644
--- a/src/views/Rules/Rule/index.jsx
+++ b/src/views/Rules/Rule/index.jsx
@@ -94,12 +94,12 @@ export default function Rule({ isNewRule, ...props }) {
   const [updateSCAction, updateSC] = useAction(updateScheduledChange);
   const [deleteSCAction, deleteSC] = useAction(deleteScheduledChange);
   const isLoading =
-    fetchRuleAction.loading ||
-    scheduledChangeAction.loading ||
-    products.loading ||
-    channels.loading;
+    fetchRuleAction.loading || products.loading || channels.loading;
   const actionLoading =
-    addSCAction.loading || updateSCAction.loading || deleteSCAction.loading;
+    addSCAction.loading ||
+    updateSCAction.loading ||
+    deleteSCAction.loading ||
+    scheduledChangeAction.loading;
   const error =
     fetchRuleAction.error ||
     scheduledChangeAction.error ||
@@ -543,10 +543,11 @@ export default function Rule({ isNewRule, ...props }) {
           </Grid>
         </Fragment>
       )}
-      {!isLoading && !actionLoading && (
+      {!isLoading && (
         <Fragment>
           <Tooltip title={isNewRule ? 'Create Rule' : 'Update Rule'}>
             <Fab
+              disabled={actionLoading}
               onClick={isNewRule ? handleCreateRule : handleUpdateRule}
               color="primary"
               className={classNames({
@@ -559,7 +560,7 @@ export default function Rule({ isNewRule, ...props }) {
           {hasScheduledChange && (
             <SpeedDial ariaLabel="Secondary Actions">
               <SpeedDialAction
-                disabled={scheduledChangeAction.loading}
+                disabled={actionLoading}
                 icon={<DeleteIcon />}
                 tooltipOpen
                 tooltipTitle="Cancel Pending Change"
```

src/views/Rules/Rule/index.jsx
mozbhearsum: I think it's best to keep these lowercase actually, as they end up in the update XML (eg: https://aus5.mozilla.org/update/3/Firefox/33.0/20141202185629/Darwin_x86_64-gcc3-u-i386-x86_64/en-US/release/default/default/default/update.xml). (I realize we can have the label and value set differently, I just think it might be confusing/surprising to people).

src/views/Rules/Rule/index.jsx
mozbhearsum: Ah, to disable the controls instead of hide them? I was wondering about that :)

helfi92: Yeah :)

Issue #58, opened 2019-07-08
implement view for showing releases


mozbhearsum:

helfi92:
Planning to work on this next. As always, let me know if I should focus my energy elsewhere for now.

mozbhearsum:
Makes sense to me. Once this, #63, and #64 are done we'll have all of the major pieces for an MVP and it'll be easier to find all of the small issues that we need to resolve before cutting over.

helfi92:
> Once this, #63, and #64 are done we'll have all of the major pieces for an MVP and it'll be easier to find all of the small issues that we need to resolve before cutting over.

Exciting!! 🎉

helfi92:
Resolved in https://github.com/mozilla-frontend-infra/balrog-ui/commit/7b61f42ee99ccb6a32db2c8a838d4a29c8bb87dc.

PR #74, opened 2019-07-10
Add initial layout for the releases view

helfi92:
Relates to https://github.com/mozilla-frontend-infra/balrog-ui/issues/58.

### DIFF
--- a/src/components/Dashboard/index.jsx
+++ b/src/components/Dashboard/index.jsx
@@ -9,11 +9,11 @@ import menuItems from './menuItems';
 import Link from '../../utils/Link';
 import UserMenu from './UserMenu';
 import SettingsMenu from './SettingsMenu';
-import { CONTENT_MAX_WIDTH } from '../../utils/constants';
+import { CONTENT_MAX_WIDTH, APP_BAR_HEIGHT } from '../../utils/constants';
 
 const useStyles = makeStyles(theme => ({
   appbar: {
-    height: theme.spacing(8),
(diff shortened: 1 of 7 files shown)


helfi92:
> Home/end/pageup/pagedown don't behave as expected - they barely move the page at all.

What is "Home" and "end". Are those buttons on your keyboard?

> Noticed this exception in the console:

I'll send a patch here.

> A text field to filter by name -- this is crucial because of how many Releases we have.

Agreed. I will create a `Search` component in a follow-up and integrate it in this view.

> Hide/Disable the "Delete" button for Releases that are used by one or more Rules.

I'll send a patch here.

> Add a button for flipping the read-only/read-write flag.

Sounds good. I might tackle this in a follow-up.

> Handle scheduled changes.

We can take of that in a follow-up.

mozbhearsum:
> What is "Home" and "end". Are those buttons on your keyboard?

Yes :).

helfi92:
> Yes :).

Ah :( I figured out the issue but don't have a solution on top of my head. I will investigate more. Btw, this issue is probably also on the rules view as well now that the jump to hash has been merged.

mozbhearsum:
> > Yes :).
> 
> Ah :( I figured out the issue but don't have a solution on top of my head. I will investigate more. Btw, this issue is probably also on the rules view as well now that the jump to hash has been merged.

Ah, indeed...

helfi92:
To be done later:

* Add a button for flipping the read-only/read-write flag
* Handle scheduled changes

PR #75, opened 2019-07-11
Fix scrolling issue with Rules view

helfi92:


helfi92:
I assume it's coming from the react-virtualized library now that we are no longer using react-window but I'm not sure. I think we can ignore for now.

PR #79, opened 2019-07-11
Allow flipping read-only/read-write

helfi92:


helfi92:
Rebased.

src/components/ReleaseCard/index.jsx
```diff
@@ -95,32 +101,35 @@ function ReleaseCard(props) {
...
-          </Tooltip>
+          <div>
+            <FormControlLabel
+              control={
+                <Switch
+                  checked={release.read_only}
+                  onChange={handleAccessChange}
+                />
+              }
+              label="Read only"
```
mozbhearsum: I think we should change this text somehow. `Read only` sounds like a statement to me (as in, this release is read only, even if it's unchecked). If the label text changed with the state that would work, or if it was more clearly a label. I'm open to other ideas too.

src/views/Releases/ListReleases/index.jsx
```diff
@@ -64,12 +67,28 @@ function ListPermissions(props) {
     }
   }, [hash, filteredReleases]);
 
+  const handleAccessChange = ({ release, checked }) => {
+    setDialogState({
+      open: true,
+      title: checked ? 'Read/Write?' : 'Read Only?',
```
mozbhearsum: I think this is backwards. When I click the button for CDM-15, which is currently read-write, the title of the dialog is `Read/Write?`

src/views/Releases/ListReleases/index.jsx
```diff
@@ -64,12 +67,28 @@ function ListPermissions(props) {
...
+      confirmText: 'Yes',
+      body: `This would make the permissions of release ${release.name} ${
```
mozbhearsum: Let's avoid using the word `permissions` here, because it overlaps with the idea of user permissions. How about `This would make ${release.name} writable/read only`?

src/components/ReleaseCard/index.jsx
helfi92: > or if it was more clearly a label.

Perhaps something like:

Screen Shot 2019-07-12 at 11 03 40 AM

helfi92: I picked a switch because a switch usually means to toggle the state of a setting. Once toggled to the left, the actions of that card become clickable.

helfi92: Let me know if you still don't like the current approach. I can try to think of another alternative.

mozbhearsum: Yeah, I like the switch control, the text just threw me a bit. I like the screenshot above. Grouping the switch and the text together clears up the ambiguity I had.

PR #81, opened 2019-07-12
Use dotenv for enviroment specific config

mozbhearsum:
I'm getting ready to deploy to our stage environment and we need to be able to override the backend url + auth0 config at build time. My intention is to drop in a `.env` with the appropriate values before building/publishing as part of the work going on in https://github.com/mozilla/balrog/pull/947.


.neutrinorc.js
```diff
@@ -26,13 +28,13 @@ module.exports = {
...
-          AUTH0_RESPONSE_TYPE: 'token id_token',
-          AUTH0_SCOPE: 'full-user-credentials openid profile email',
-          AUTH0_REDIRECT_URI: `http://localhost:${port}/login`,
+          BALROG_ROOT_URL: process.env.BALROG_ROOT_URL || 'https://localhost:8010',
+          AUTH0_CLIENT_ID: process.env.AUTH0_CLIENT_ID || 'GlZhJQfx52b7MLQ19AjuTJHieiB4oh1j',
+          AUTH0_DOMAIN: process.env.AUTH0_DOMAIN || 'balrog-localdev.auth0.com',
+          AUTH0_AUDIENCE: process.env.AUTH0_AUDIENCE || 'balrog-localdev',
+          AUTH0_RESPONSE_TYPE: process.env.AUTH0_RESPONSE_TYPE || 'token id_token',
+          AUTH0_SCOPE: process.env.AUTH0_SCOPE || 'full-user-credentials openid profile email',
+          AUTH0_REDIRECT_URI: process.env.AUTH0_REDIRECT_URI || `http://localhost:${port}/login`,
```
helfi92: The values specified here are the default ones. Once `dotenv` injects variables before this file runs, all of these values will be ignored if they are defined in the `.env` file. That being said, I think we can remove `process.env.[..]` from here.

.neutrinorc.js
```diff
@@ -1,3 +1,5 @@
+require('dotenv').config()
```
helfi92: We should be able to use the CLI directly:

```json
"build": "dotenv webpack -- --mode production",
"start": "dotenv webpack-dev-server -- --mode development",
```

.neutrinorc.js
mozbhearsum: Isn't this what we want? For local development, what's in the repo is fine (we'll use the defaults specified above). When building for stage or prod, we'll drop in a `.env`, which will override them.

helfi92: By the time `.neutrinorc.js` runs, the environment variables would have already been set. Neutrino will ignore all of the key value pairs that are already defined.

helfi92: You could technically just have:

```js
{
  env: [
    'AUTH0_AUDIENCE',
    'AUTH0_RESPONSE_TYPE',
    // ...
  ],
};
```

Neutrino will throw a warning if it notices that any of these values are not already defined in the environment. https://master.neutrinojs.org/packages/web/#environment-variables

mozbhearsum: Ah, I didn't realize Neutrino would automatically look in the environment for overrides...that makes things simpler!

mozbhearsum: Given that, it looks like this file doesn't need to change at all...

package.json
```diff
@@ -23,6 +23,8 @@
     "classnames": "^2.2.6",
     "clipboard-copy": "^3.0.0",
     "date-fns": "2.0.0-alpha.37",
+    "dotenv": "^8.0.0",
+    "dotenv-cli": "^2.0.0",
```
helfi92: These are not required in the codebase so we could make them dev dependencies.

.neutrinorc.js
helfi92: Indeed :)

package.json
mozbhearsum: Ah, of course. Let's do that.

PR #84, opened 2019-07-16
Deep link scheduled changes in a rule card

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/71.

### DIFF
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -33,7 +33,7 @@ import getDiffedProperties from '../../utils/getDiffedProperties';
 
 const useStyles = makeStyles(theme => ({
   root: {
-    '& h2': {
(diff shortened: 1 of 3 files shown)


src/components/RuleCard/index.jsx
```diff
@@ -662,8 +662,15 @@ function RuleCard({ rule, onRuleDelete, ...props }) {
             <div className={classes.scheduledChangesHeader}>
               <Typography
                 className={classes.scheduledChangesTitle}
+                component="h4"
```
mozbhearsum: Out of curiosity, why use an `h4` here, but an `h2` for other Rules?

helfi92: In the UI the rule name has a bigger font and a bold weight. It is purely semantics.

PR #87, opened 2019-07-17
Implement release deletions

mozbhearsum:
Pretty much the same as we did for Rules, except I had to add a second `DialogAction` since we already had one for read only on this view.

### DIFF
--- a/src/views/Releases/ListReleases/index.jsx
+++ b/src/views/Releases/ListReleases/index.jsx
@@ -97,16 +97,16 @@ function ListReleases(props) {
 
   // TODO Add mutation
   const handleReadOnlySubmit = () => {};
-  const handleReadOnlyClose = () => {
+  const handleReadOnlyClose = state => {
     setDialogState({
-      ...dialogState,
+      ...state,
       open: false,
     });
   };
 
-  const handleReadOnlyError = error => {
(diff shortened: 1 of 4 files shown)


src/views/Releases/ListReleases/index.jsx
```diff
@@ -149,16 +227,28 @@ function ListPermissions(props) {
...
+        title={deleteDialogState.title}
+        body={deleteDialogState.body}
+        open={deleteDialogState.open}
+        error={deleteDialogState.error}
+        confirmText={deleteDialogState.confirmText}
+        onSubmit={handleDeleteSubmit}
+        onClose={handleDeleteClose}
+        onError={handleDeleteError}
+        onComplete={handleDeleteComplete}
       />
```
helfi92: I think we should be able to handle both actions in a single dialog. We would need to add `onSubmit`, `onClose`, `onError`, `onComplete` to the initial dialog state and set them something like `Function.prototype` or `() => {}`.

mozbhearsum:
I was thinking that might be related to the issue with dialogState not working as expected, but maybe not. I'll dig into it more.

helfi92:
I tried to debug the issue with `dialogState` but couldn't find the culprit. I created a playground from scratch to see if this was a bug with React but it doesn't appear to be so.

PR #96, opened 2019-07-18
Support basic error message if the backend cannot be contacted.

mozbhearsum:
Ideally, we'd block rendering of pages on this (to avoid double errors such as below), but I couldn't figure out how to make that work (blocking the entire render method means nothing at all renders, of course...)

image

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,10 +1,9 @@
 import { hot } from 'react-hot-loader/root';
-import React, { Fragment, useState, useEffect } from 'react';
+import React, { Fragment, useState } from 'react';
 import { Authorize } from 'react-auth0-components';
 import CssBaseline from '@material-ui/core/CssBaseline';
 import { ThemeProvider } from '@material-ui/styles';
 import axios from 'axios';
-import ErrorPanel from './components/ErrorPanel';
 import { AuthContext } from './utils/AuthContext';
 import { BASE_URL, USER_SESSION } from './utils/constants';
 import theme from './theme';
@@ -66,7 +65,6 @@ const App = () => {
   // Wait until authorization is done before rendering
   // to make sure users who are logged in are able to access protected views
   const [ready, setReady] = useState(Boolean(!session));
-  const [backendError, setBackendError] = useState('');
   const handleAuthorize = user => {
     setAuthContext({
       ...authContext,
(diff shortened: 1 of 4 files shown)


src/App.jsx
```diff
@@ -95,6 +109,7 @@ const App = () => {
       <CssBaseline />
       <AuthContext.Provider value={authContext}>
         <ThemeProvider theme={theme}>
+          {backendError && <ErrorPanel fixed error={backendError} />}
```
helfi92: I think we can implement the logic in `Main.jsx` so that we are able to use the theme object from material UI. Also, in order to only show the backend error, we could do something like that in `Main.jsx`:

```jsx
// ...

backendError: {
    marginTop: // add some margin top
},

// ...

return backendError ? (
 // <ErrorPanel fixed className={classes.backendError} error={...} />
) : (
 // 
);
```

mozbhearsum: I ended up going a bit further here and disabling most of the UI when the backend is unavailable.

src/App.jsx
```diff
@@ -32,8 +32,9 @@ axios.interceptors.request.use(config => {
 axios.interceptors.response.use(
   response => response,
   error => {
-    const errorMsg =
-      error.response.data.exception || error.response.data.detail || null;
+    const errorMsg = error.response
+      ? error.response.data.exception || error.response.data.detail || null
+      : error.message;
```
helfi92: 👌

mozbhearsum:
> One thing is that I am able to see the error when I open the balrog UI in Chrome. Should the error also say that it only works on Firefox?

Is that true? I've used Chrome to debug a few times and haven't seen any issues.

mozbhearsum:
> > One thing is that I am able to see the error when I open the balrog UI in Chrome. Should the error also say that it only works on Firefox?
> 
> Is that true? I've used Chrome to debug a few times and haven't seen any issues.

I just realized that you meant that you always see this error on Chrome. This is because of the self signed SSL certificate we use for local development. If you visit https://localhost:8010/ directly and accept it, then the UI should work.

helfi92:
> This is because of the self signed SSL certificate we use for local development. If you visit https://localhost:8010/ directly and accept it, then the UI should work.

Ah. That fixed my issue. Many thanks :)

PR #101, opened 2019-07-19
Show pending rule signoffs (fixes #61)

### LINKED ISSUE #61: display required & pending signoffs on rule scheduled changes

mozbhearsum:


src/components/RuleCard/index.jsx
```diff
@@ -707,6 +711,15 @@ function RuleCard({ rule, onRuleDelete, ...props }) {
             )}
           </Fragment>
         )}
+        {requiresSignoff && (
+          <Fragment>
+            <Divider className={classes.divider} />
```
helfi92: We probably don't need a divider line between scheduled changes and requires signoffs.

mozbhearsum:
> I don't seem to see values for Requires Signoffs From

Hm, so what's happening here is that there's a signoff even though the change doesn't require signoff (this is done by the backend when the change is scheduled). Now that we're using scheduled changes as the only way to modify rules, we should probably make sure we don't show this section unless signoff is required, instead of just checking for the prescence of a scheduled change.

helfi92:
> Now that we're using scheduled changes as the only way to modify rules, we should probably make sure we don't show this section unless signoff is required, instead of just checking for the prescence of a scheduled change.

Makes sense.

src/components/RuleCard/index.jsx
```diff
@@ -682,6 +698,15 @@ function RuleCard({ rule, onRuleDelete, readOnly, ...props }) {
             )}
           </Fragment>
         )}
+        {requiresSignoff && (
+          <Fragment>
+            <div className={classes.space} />
```
helfi92: Instead of adding an empty div with a class, allow the SignoffSummary to handle a `className`:

```jsx
// SignoffSummary/index.jsx

import classNames from 'classnames';
import { object } from 'prop-types';

// ...

function SignoffSummary(props) {
  const { requiredSignoffs, signoffs, className } = props;

  // ...

  <div className={classNames(classes.listWrapper, className)}>
    // ...
  </div>
}


SignoffSummary.propTypes = {
  // ...
  className: object,
};

SignoffSummary.defaultProps = {
  className: null,
};
```

src/components/RuleCard/index.jsx
```diff
@@ -48,6 +50,9 @@ const useStyles = makeStyles(theme => ({
       },
     },
   },
+  space: {
+    paddingTop: theme.spacing(1),
```
helfi92: `paddingTop: theme.spacing(2),`

src/components/SignoffSummary/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
+import ListSubheader from '@material-ui/core/ListSubheader';
+import Typography from '@material-ui/core/Typography';
+import { object } from 'prop-types';
+
+const useStyles = makeStyles(theme => ({
+  listSubheader: {
+    lineHeight: 1.5,
+    marginBottom: theme.spacing(1),
```
helfi92: `marginBottom: theme.spacing(0.5),`

Also let add `paddingLeft: theme.spacing(1),`.

src/components/SignoffSummary/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+            Requires Signoffs From
+          </ListSubheader>
+        }>
+        {Object.entries(requiredSignoffs).map(([role, count], index) => {
+          const key = `${role}-${index}`;
+
+          return (
+            <ListItem key={key} dense className={classes.requiresSignoffsList}>
+              <ListItemText
+                primary={`${count} member${count > 1 ? 's' : ''} of ${role}`}
```
helfi92: Let's set top and bottom margins of both ListItemText component to 0.

src/components/SignoffSummary/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+        subheader={
+          <ListSubheader className={classes.listSubheader}>
+            Requires Signoffs From
+          </ListSubheader>
+        }>
+        {Object.entries(requiredSignoffs).map(([role, count], index) => {
+          const key = `${role}-${index}`;
+
+          return (
+            <ListItem key={key} dense className={classes.requiresSignoffsList}>
```
helfi92: Remove the `dense` prop from both `ListItem` components. Instead add `paddingTop: 0, paddingBottom: 0` to the `requiresSignoffsList` class.

src/components/SignoffSummary/index.jsx
```diff
@@ -0,0 +1,84 @@
...
+    </div>
+  );
+}
+
+SignoffSummary.propTypes = {
+  requiredSignoffs: object.isRequired,
+  signoffs: object.isRequired,
+};
+
+export default SignoffSummary;
```
helfi92: The styling changes suggested above should hopefully do the following change:

#### Before:

Screen Shot 2019-07-23 at 9 26 36 AM


#### After:

Screen Shot 2019-07-23 at 9 20 11 AM

src/components/RuleCard/index.jsx
```diff
@@ -682,6 +698,15 @@ function RuleCard({ rule, onRuleDelete, readOnly, ...props }) {
             )}
           </Fragment>
         )}
+        {requiresSignoff && (
+          <Fragment>
+            <div className={classes.space} />
+            <SignoffSummary
```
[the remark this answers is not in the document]
mozbhearsum: Effectively, this would mean we hide them for the history view, which I think makes sense.

src/components/SignoffSummary/index.jsx
mozbhearsum: Thanks for this, I see that it cleans up the alignment!

src/components/SignoffSummary/index.jsx
mozbhearsum: This caused a bit of inconsistency in the horizontally alignment. I ended wrapping the Required Signoffs text in a `<Typography>` to fix it.

src/components/SignoffSummary/index.jsx
```diff
@@ -38,9 +46,14 @@ function SignoffSummary(props) {
...
-            <ListItem key={key} dense className={classes.requiresSignoffsList}>
+            <ListItem key={key} className={classes.signoffsList}>
               <ListItemText
-                primary={`${count} member${count > 1 ? 's' : ''} of ${role}`}
+                primary={
+                  <Typography>{`${count} member${
+                    count > 1 ? 's' : ''
+                  } of ${role}`}</Typography>
+                }
+                className={classes.listItemText}
```
helfi92: Looks a bit off with the Typography component since the variant is changed:

#### With Typogpraphy

Screen Shot 2019-07-23 at 10 26 37 AM


#### Without Typography

Screen Shot 2019-07-23 at 10 26 20 AM

`ListItemText` internally uses a Typography component to render the primary prop.

src/views/Rules/ListRules/index.jsx
```diff
@@ -415,6 +415,28 @@ function ListRules(props) {
...
+
+          // Signoffs and required roles are beside one another, so if there's
+          // the same number or fewer signoffs than required roles, we don't
+          // need to add extra space for them. If the signoffs are more
+          // mnumerous than the number of required roles, we need more space
+          // for the extras.
+          if (nSignoffs > requiredRoles) {
+            height += theme.spacing(5 * (nSignoffs - requiredRoles));
+          }
+        }
```
helfi92: I think the heights need to be updated. Navigating to http://localhost:9000/rules#scId=21 shows cards with no spaces in between :



Screen Shot 2019-07-23 at 10 28 43 AM

src/components/SignoffSummary/index.jsx
mozbhearsum: Hm, it looks a little different for me. With the Typography, I see:
image

src/views/Rules/ListRules/index.jsx
mozbhearsum: Hm, indeed. Once we figure out the other issue I'll take a look at this.

src/components/SignoffSummary/index.jsx
helfi92: Interesting. Let's try the same thing but written differently. Can you try:

```diff
diff --git a/src/components/SignoffSummary/index.jsx b/src/components/SignoffSummary/index.jsx
index ccb9b7b..998734f 100644
--- a/src/components/SignoffSummary/index.jsx
+++ b/src/components/SignoffSummary/index.jsx
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import { object } from 'prop-types';
 import classNames from 'classnames';
 import { makeStyles } from '@material-ui/styles';
@@ -26,6 +26,10 @@ const useStyles = makeStyles(theme => ({
     paddingTop: 0,
     paddingBottom: 0,
   },
+  signedBy: {
+    display: 'flex',
+    alignItems: 'center',
+  },
 }));
 
 function SignoffSummary(props) {
@@ -48,11 +52,7 @@ function SignoffSummary(props) {
           return (
             <ListItem key={key} className={classes.signoffsList}>
               <ListItemText
-                primary={
-                  <Typography>{`${count} member${
-                    count > 1 ? 's' : ''
-                  } of ${role}`}</Typography>
-                }
+                primary={`${count} member${count > 1 ? 's' : ''} of ${role}`}
                 className={classes.listItemText}
               />
             </ListItem>
@@ -71,14 +71,21 @@ function SignoffSummary(props) {
             {listOfSignoffs.map(([username, signoffRole]) => (
               <ListItemText
                 key={username}
+                disableTypography
                 primary={
-                  <Fragment>
+                  <Typography
+                    component="p"
+                    className={classes.signedBy}
+                    variant="body2">
                     {username}
-                    {' - '}
-                    <Typography color="textSecondary" variant="caption">
+                    &nbsp; - &nbsp;
+                    <Typography
+                      component="span"
+                      color="textSecondary"
+                      variant="caption">
                       {signoffRole}
                     </Typography>
-                  </Fragment>
+                  </Typography>
                 }
                 className={classes.listItemText}
               />
```

mozbhearsum: That changes the padding ever so slightly, but still looks good here:
image

helfi92: Looks good to me.

helfi92:
A rebase would be appreciated :)

mozbhearsum:
> A rebase would be appreciated :)

Yeah, I was just doing that :). And there's more spacing issues to address after doing so.

helfi92:
Ah, apologies. I should have waited for a "request to review" before jumping in.

mozbhearsum:
Okay, this is ready again :). I ended up adjusting the main row padding on the RuleCards as well, because after fixing the jaws/mig64 display I noticed that we end up with a LOT of extra space between cards. If the fractional spacing is uncouth we could switch to `theme.padding(7)` without it being too extreme.

helfi92:
The cards are a bit off when I pull down this branch. Not sure what's happening. You could try rebasing on master to see if that fixes it.

Screen Shot 2019-07-25 at 8 47 09 AM

mozbhearsum:
Hm, I don't see that myself. I'm guessing this is because `jaws` and `mig64` are undefined on your side, which is causing them to be shown on the Cards, but no space is calculated for them (I just noticed that we check for null and undefined when calculating height).

I'm curious how they get set to undefined for you but not me. In any case, the best fix is probably to check for null and undefined in RuleCard, too.

src/views/Rules/ListRules/index.jsx
```diff
@@ -416,6 +417,35 @@ function ListRules(props) {
...
         }
+
+        if (Object.keys(rule.scheduledChange.required_signoffs).length > 0) {
+          const requiredRoles = Object.keys(
+            rule.scheduledChange.required_signoffs
+          ).length;
+          const nSignoffs = Object.keys(rule.scheduledChange.signoffs).length;
+
+          // Add space for the padding above the summary,
+          // and at the bottom of the section
```
helfi92: Which summary. The comment?

src/views/Rules/ListRules/index.jsx
```diff
@@ -391,7 +391,8 @@ function ListRules(props) {
         keys.filter(key => thirdColumn.includes(key)).length
       );
 
-      height += rows * theme.spacing(8);
+      // Each row needs 50px, 50 / 8 = 6.25
+      height += rows * theme.spacing(6.25);
```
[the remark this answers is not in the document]
mozbhearsum: I'm having trouble screenshotting the highlighted field, but you're right -- it's 50px for me.

src/views/Rules/ListRules/index.jsx
```diff
@@ -416,6 +417,35 @@ function ListRules(props) {
...
+          height += theme.spacing(2);
+
+          // Add space for the "Requires Signoff From" title
+          height += theme.spacing(3);
+
+          // Add space for each row of required roles
+          // 2 is actually a little less than needed, but we rarely
+          // require more than 2 signoffs, and erring low looks better
+          // than erring high.
+          height += requiredRoles * theme.spacing(2);
```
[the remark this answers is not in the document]
mozbhearsum: > I think we could do Math.max(nSignoffs, requiredRoles) and remove the extra logic below.

Good point, will do.

> Also, on my end the row height is a bit over 20px which means theme.spacing(2) may be not be enough. Are you seeing a different row height on your end?

Yup, I've got ~18.5px here.

src/views/Rules/ListRules/index.jsx
mozbhearsum: Summary meaning SignoffSummary. This is to account for the padding that we're passing to it via className.

src/components/RuleCard/index.jsx
```diff
@@ -568,7 +584,7 @@ function RuleCard({ rule, onRuleDelete, readOnly, ...props }) {
                   />
                 </ListItem>
               )}
-              {rule.jaws && (
+              {rule.jaws != null && (
                 <ListItem className={classes.listItem}>
                   <ListItemText
                     primaryTypographyProps={{
```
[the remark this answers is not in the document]
mozbhearsum: I'm still not sure how you're seeing it in the primary area of a `RuleCard`. As far as I know (and in my own local testing) we don't show that section for scheduled inserts....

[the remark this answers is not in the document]
helfi92: I think it was showing before https://github.com/mozilla-frontend-infra/balrog-ui/pull/101/commits/41dd2330be5676087892527b1c41613df7d75df2 but not after.

src/views/Rules/ListRules/index.jsx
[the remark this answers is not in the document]
helfi92: Interesting! Ok, I will try to investigate to see if there's anything we can do about it. Worst case we just support the highest height (i.e., `theme.spacing(3)` instead of `theme.spacing(2)`).

src/views/Rules/ListRules/index.jsx
helfi92: Ah okay. I understand now. Makes sense.

src/views/Rules/ListRules/index.jsx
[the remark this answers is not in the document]
mozbhearsum: Yeah, we can do that. It's slightly worse looking on my screen, but not dealbreaking.

helfi92:
Do you mind performing a rebase and checking if the gap in https://github.com/mozilla-frontend-infra/balrog-ui/pull/117#issuecomment-515528327 is fixed, please. Also, the height will now need to use  the [`elementsHeight`](https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/utils/elementsHeight.js) utility to get an exact height.

mozbhearsum:
> Do you mind performing a rebase and checking if the gap in [#117 (comment)](https://github.com/mozilla-frontend-infra/balrog-ui/pull/117#issuecomment-515528327) is fixed, please. Also, the height will now need to use the [`elementsHeight`](https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/utils/elementsHeight.js) utility to get an exact height.

Will do.

mozbhearsum:
Let me know if I used the new helpers correctly. I dug into the underlying components/elements to try to figure out the correct ones to use.

While finishing this up I did notice that there's still a small inconsistency (even on master) that appears to be caused by the diff viewer. I dug into a bit and couldn't find anything in the spacing that was mismatched with what I saw in the Inspector, though. It's not bad enough that I'm concerned about it, just wanted to mention it in case you saw it.
image

src/views/Rules/ListRules/index.jsx
```diff
@@ -433,6 +433,25 @@ function ListRules(props) {
...
+          ).length;
+          const nSignoffs = Object.keys(rule.scheduledChange.signoffs).length;
+          // Required Roles and Signoffs are beside one another, so we only
+          // need to account for the one with the most items.
+          const signoffRows = Math.max(requiredRoles, nSignoffs);
+
+          // Padding above the summary
+          height += theme.spacing(2);
+
+          // The "Requires Signoff From" title and the padding beneath it
```
helfi92: `margin beneath it`

src/views/Rules/ListRules/index.jsx
```diff
@@ -433,6 +433,25 @@ function ListRules(props) {
...
+          height += body1TextHeight() + theme.spacing(0.5);
```
helfi92: The subheader font-size is `0.875rem` which corresponds to the size of `body2TextHeight`.

src/views/Rules/ListRules/index.jsx
```diff
@@ -433,6 +433,25 @@ function ListRules(props) {
...
+
+          // Space for however many rows exist.
+          height += signoffRows * (body2TextHeight() + theme.spacing(1));
```
helfi92: Rows don't seem to have margins between them so we could probably remove `theme.spacing(1)`.

mozbhearsum:
> It might be because of the "Comment" section since we set the height to account fora maximum of 3 lines even though most cards either have no comments or have a comment that fits on a single line. Because of that, we see some minimal discrepancies in spacing.

Ah! Of course. I suppose if it really ends up bothering us, we could try to make a better estimate of spacing based on the length of the comment - but I don't think it's worth doing anything about now.

src/views/Rules/ListRules/index.jsx
mozbhearsum: Hm, OK.

Issue #63, opened 2019-07-08
UI for rules history


mozbhearsum:
At minimum, we need to be able to view history by rule id. It would be even better if we could look at the entire Rules table at a specific point in time, though (this would probably need some backend support).

helfi92:
![Screen Shot 2019-07-22 at 11 03 55 AM](https://user-images.githubusercontent.com/3766511/61643241-72920e80-ac70-11e9-88c7-1316a8d79735.png)

PR #102, opened 2019-07-22
Add revision view for rules

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/63.

Screen Shot 2019-07-22 at 11 03 55 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@date-io/date-fns": "^1.3.7",
     "@hot-loader/react-dom": "^16.8.6",
-    "@material-ui/core": "^4.1.3",
+    "@material-ui/core": "^4.2.1",
     "@material-ui/icons": "^4.2.1",
     "@material-ui/lab": "^4.0.0-alpha.11",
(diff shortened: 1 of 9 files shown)


helfi92:
> It looks like not all revisions are displayed due to not handling pagination. The revisions endpoints support page & limit. For Rules, it's probably easiest just to set limit to a Really High number (10,000 ought to be enough for the forseeable future).

@mozbhearsum How often do rules have revisions > 100? If it's often, we should probably look into virtualizing the list.

mozbhearsum:
> > It looks like not all revisions are displayed due to not handling pagination. The revisions endpoints support page & limit. For Rules, it's probably easiest just to set limit to a Really High number (10,000 ought to be enough for the forseeable future).
> 
> @mozbhearsum How often do rules have revisions > 100? If it's often, we should probably look into virtualizing the list.

It's pretty common for Rules that we'd be interested in history of, so it sounds like we should do this.

helfi92:
Sounds good. I guess we can do this virtualization in a follow-up.

PR #109, opened 2019-07-23
Allow users to signoff on changes to required signoffs

mozbhearsum:
This allows users to signoff and revoke signoffs on changes to required signoffs. If a user holds only a single Role, the signoff is a single click. If they hold multiple roles, we need to ask which they want to signoff with (they're only allowed to signoff under one Role).

The fact that a signoff can happen immediately or after going through a Dialog made things a little bit awkward, but I did my best to make sure state updates and errors were done and shown in the right places (ie: don't update state until handleDialogComplete fires). The updates to state themselves are a little bit awkward (cloning the entire object seems nasty), but I couldn't figure out a nicer way to do it given the data structure.

I intend to carry this work over to Rules, Releases, and User Permission signoffs too.


src/components/SignoffCardEntry/index.jsx
```diff
@@ -66,7 +66,7 @@ function getStatus(entry) {
 
 function SignoffCardEntry(props) {
   const classes = useStyles();
-  const { user, entry, name } = props;
+  const { user, entry, name, onSignoff, onRevoke } = props;
```
helfi92: Let's add these to `propTypes`.

src/views/RequiredSignoffs/ListSignoffs/index.jsx
```diff
@@ -48,12 +63,28 @@ const useStyles = makeStyles(theme => ({
...
+  const [signoffAction, signoff] = useAction(signoffRequiredSignoff);
+  const [revokeAction, revoke] = useAction(revokeRequiredSignoff);
+  const [rolesAction, getRoles] = useAction(getUserInfo);
+  const loading = getRSAction.loading || rolesAction.loading;
+  const error =
+    getRSAction.error ||
+    revokeAction.error ||
+    rolesAction.error ||
+    // If there's more than one role, this error is shown inside of the dialog
+    (roles.length === 1 && signoffAction.error);
```
helfi92: 👌

src/components/SignoffCardEntry/index.jsx
helfi92: Note: If they are not required, then we should probably define a default value in `defaultProps`. Setting them to `null` should be good enough.

mozbhearsum: Ah, I should actually make them required - my bad.

Issue #103, opened 2019-07-22
Button with link in Chrome only


helfi92:
<img width="391" alt="Screen Shot 2019-07-22 at 11 14 01 AM" src="https://user-images.githubusercontent.com/3766511/61643955-dcf77e80-ac71-11e9-8dd1-3f784d330414.png">

Duplicate and Update should not be underlined.

PR #110, opened 2019-07-23
Add theme

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/107 and https://github.com/mozilla-frontend-infra/balrog-ui/issues/103.


src/theme.js
```diff
@@ -8,15 +9,18 @@ const SPACING = {
 };
 
 export default createMuiTheme({
+  palette: {
+    primary: deepPurple,
```
mozbhearsum: How do you feel about pulling one of the shades of purple from https://github.com/mozilla/balrog/blob/master/ui/app/img/balrog.svg for the primary colour? Or will that make the secondary/error colours more difficult to match and/or create other issues?

src/components/Button/index.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
mozbhearsum: Should we be replacing all of our usage of Button and IconButton with this new one?

helfi92: We can replace our usages of `Button` but not `IconButton` since it's not based from the Button component. I'll do the change.

src/theme.js
helfi92: Sure thing. I'll give it a try.

Issue #100, opened 2019-07-19
allow users to make/revoke signoffs on required signoffs


mozbhearsum:

mozbhearsum:
Fixed in #109.

PR #116, opened 2019-07-25
Allow users to sign off on rule scheduled changes (fixes #62)

### LINKED ISSUE #62: allow users to signoff and revoke signoffs on rule scheduled changes
If a User only holds one role, we can assume they will signoff with it. If they hold multiple roles, we need to prompt them to choose one to use.

mozbhearsum:
This is pretty much copy paste from #100, except for:
* I realized that the wrappers for making signoff/revoke requests were already generic enough to handle other objects, so I moved them to a better location.
* The dialog handling code is a bit more complicated because we're re-using the existing dialog from the view, like we did on the Releases view.

And it's built on top of #101.

### DIFF
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -718,35 +718,51 @@ function RuleCard({
       </CardContent>
       {!readOnly && (
         <CardActions className={classes.cardActions}>
-          <Link
-            className={classes.link}
-            to={{
-              pathname: '/rules/create',
-              state: {
-                rule,
-              },
-            }}>
-            <Button color="secondary">Duplicate</Button>
-          </Link>
-          <Link
-            className={classes.link}
-            to={
-              rule.rule_id
-                ? `/rules/${rule.rule_id}`
-                : `/rules/create/${rule.scheduledChange.sc_id}`
-            }>
-            <Button color="secondary">Update</Button>
-          </Link>
-          <Button color="secondary" onClick={() => onRuleDelete(rule)}>
+          {user ? (
+            <Link
+              className={classes.link}
+              to={{
+                pathname: '/rules/create',
+                state: {
+                  rule,
+                },
+              }}>
+              <Button color="secondary">Duplicate</Button>
+            </Link>
+          ) : (
+            <Button color="secondary" disabled>
+              Duplicate
+            </Button>
+          )}
+          {user ? (
+            <Link
+              className={classes.link}
+              to={
+                rule.rule_id
+                  ? `/rules/${rule.rule_id}`
+                  : `/rules/create/${rule.scheduledChange.sc_id}`
+              }>
+              <Button color="secondary">Update</Button>
+            </Link>
+          ) : (
+            <Button color="secondary" disabled>
+              Update
+            </Button>
+          )}
+          <Button
+            color="secondary"
+            disabled={!user}
+            onClick={() => onRuleDelete(rule)}>
             Delete
           </Button>
-          {requiresSignoff &&
+          {!readOnly &&
+            requiresSignoff &&
             (user && user.email in rule.scheduledChange.signoffs ? (
-              <Button color="secondary" onClick={onRevoke}>
+              <Button color="secondary" disabled={!user} onClick={onRevoke}>
                 Revoke Signoff
               </Button>
             ) : (
-              <Button color="secondary" onClick={onSignoff}>
+              <Button color="secondary" disabled={!user} onClick={onSignoff}>
                 Signoff
               </Button>
             ))}
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -751,8 +751,7 @@ function RuleCard({
             onClick={() => onRuleDelete(rule)}>
             Delete
           </Button>
-          {!readOnly &&
-            requiresSignoff &&
+          {requiresSignoff &&
             (user && user.email in rule.scheduledChange.signoffs ? (
               <Button color="secondary" disabled={!user} onClick={onRevoke}>
                 Revoke Signoff
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -140,6 +140,8 @@ const useStyles = makeStyles(theme => ({
 function RuleCard({
   rule,
   onRuleDelete,
+  onSignoff,
+  onRevoke,
   user,
   readOnly,
   onAuthorize,
@@ -737,9 +739,9 @@ function RuleCard({
           </Button>
           {requiresSignoff &&
             (user && user.email in rule.scheduledChange.signoffs ? (
-              <Button color="secondary">Revoke Signoff</Button>
+              <Button color="secondary" onClick={onRevoke}>Revoke Signoff</Button>
             ) : (
-              <Button color="secondary">Signoff as</Button>
+              <Button color="secondary" onClick={onSignoff}>Signoff as</Button>
             ))}
         </CardActions>
       )}
@@ -752,6 +754,8 @@ RuleCard.propTypes = {
   onRuleDelete: func,
   // If true, the card will hide all buttons.
   readOnly: bool,
+  onSignoff: func.isRequired,
+  onRevoke: func.isRequired,
 };
 
 RuleCard.defaultProps = {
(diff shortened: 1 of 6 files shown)


mozbhearsum:
Cancelling this review request until #101 is merged.

src/components/RuleCard/index.jsx
```diff
@@ -740,9 +742,13 @@ function RuleCard({
...
             (user && user.email in rule.scheduledChange.signoffs ? (
-              <Button color="secondary">Revoke Signoff</Button>
+              <Button color="secondary" onClick={onRevoke}>
+                Revoke Signoff
+              </Button>
             ) : (
-              <Button color="secondary">Signoff as</Button>
+              <Button color="secondary" onClick={onSignoff}>
+                Signoff
+              </Button>
```
helfi92: We should probably hide these buttons when `readOnly` is true.

src/views/Rules/ListRules/index.jsx
```diff
@@ -347,14 +356,145 @@ function ListRules(props) {
...
+  const signoffDialogBody = (
+    <FormControl component="fieldset">
+      <RadioGroup
+        aria-label="Role"
+        name="role"
+        value={signoffRole}
+        onChange={handleSignoffRoleChange}>
+        {roles.map(r => (
+          <FormControlLabel key={r} value={r} label={r} control={<Radio />} />
+        ))}
```
helfi92: I can't seem to change the selection:

Screen Shot 2019-07-30 at 9 15 58 AM

src/views/Rules/ListRules/index.jsx
```diff
@@ -347,14 +356,145 @@ function ListRules(props) {
...
+      </RadioGroup>
```
helfi92: Signing off a card makes all the cards show "Revoke Signoff". Is this expected?

Screen Shot 2019-07-30 at 9 17 24 AM

src/views/Rules/ListRules/index.jsx
mozbhearsum: Hmm, this definitely worked at one point. I'll look into it.

src/views/Rules/ListRules/index.jsx
mozbhearsum: Definitely not!

src/components/RuleCard/index.jsx
mozbhearsum: This was already the case actually (via https://github.com/mozilla-frontend-infra/balrog-ui/pull/116/files/d065ff5fa07f12dc4aae77f2f60f6a8a4f8241ce#diff-a0795e915062262d1dae44d4335bf63cR719). However, I noticed that we don't set readOnly when there's no logged in user - I've corrected that.

mozbhearsum:
Alright, all of those things should be fixed, although I'm not certain it's in the best way. The overarching issue was that the dialog body was getting copied when we opened it, which meant it wouldn't update when the page state changed.

Unrelated to the issues you found, I also fixed an issue where roles wouldn't be set if a user performed a login while already on the Rules page.

src/views/Rules/ListRules/index.jsx
```diff
@@ -476,7 +628,10 @@ function ListRules(props) {
           className={classes.ruleCard}
           key={rule.rule_id}
           rule={rule}
+          readOnly={username === ''}
```
helfi92: Let's not remove action buttons when the user is not logged in. Instead, lets disable them. Disabled action buttons will let the user know that they need to login in order to use them.

mozbhearsum:
While testing this I also noticed that because we block rendering `rolesAction.loading`, that logging in while the page is loaded causes everything to disappear while the roles are loaded. This is pretty jarring, so I removed that block to rendering.

src/views/Rules/ListRules/index.jsx
```diff
@@ -258,16 +258,18 @@ function ListRules(props) {
   }, []);
 
   useEffect(() => {
-    fetchRoles(username).then(userInfo => {
-      const roleList =
-        (userInfo.data && Object.keys(userInfo.data.data.roles)) || [];
+    if (username !== '') {
```
helfi92: If we ever we decide to change the fallback value
https://github.com/mozilla-frontend-infra/balrog-ui/blob/bb3410c401b3f20b39def7526922ef8b64a9daba/src/views/Rules/ListRules/index.jsx#L74

to something like `null` instead of `''`, then this piece of code may break. We could make this more resilient with `if (username) {`

Let me know if you had different thoughts.

src/components/RuleCard/index.jsx
```diff
@@ -712,33 +714,53 @@ function RuleCard({
...
+            </Button>
+          )}
+          <Button
+            color="secondary"
+            disabled={!user}
+            onClick={() => onRuleDelete(rule)}>
             Delete
           </Button>
-          {requiresSignoff &&
+          {!readOnly &&
```
helfi92: I just realized this check is unnecessary since we do it on top before rendering the card actions.

src/views/Rules/ListRules/index.jsx
```diff
@@ -557,17 +710,17 @@ function ListRules(props) {
...
-        body={dialogBody}
+        body={dialogMode === 'delete' ? deleteDialogBody : signoffDialogBody}
         confirmText={dialogState.confirmText}
-        onSubmit={handleDialogSubmit}
+        onSubmit={() => dialogState.handleSubmit(dialogState)}
         onError={handleDialogError}
         error={dialogState.error}
-        onComplete={handleDialogComplete}
+        onComplete={dialogState.handleComplete}
         onClose={handleDialogClose}
```
helfi92: STR:

1. Click the "Delete" button of a rule card.
2. Click cancel button in the dialog

Expected: The dialog closes smoothly
Actual: I feel like I am seeing a glimpse of the revoke signoff dialog for a bit. I see a purple button for like 200 ms.

Are you able to replicate by any chance? Apologies for not catching this earlier.

src/views/Rules/ListRules/index.jsx
mozbhearsum: Makes sense to me. I started avoiding bare `if`s after learning that they don't work as expected for Objects, but it totally makes sense here.

src/views/Rules/ListRules/index.jsx
mozbhearsum: I think I see that (although mine is briefer than ~200ms). I'm surprised though, because dialogMode isn't anything except `delete` during those STR. It continues to happen even if the dialog body is simply set to `{deleteDialogBody}` though, so I think there's something else going on here. I'll dig into it more.

mozbhearsum: Okay, so this was happening because `deleteDialogBody` ends up `null` when `handleDialogClose` fires (because its contents depend on `dialogState.item` being set). The best fix I could find was to separate setting `open: false` from resetting the rest of `dialogState`. Let me know if you have a better way.

src/views/Rules/ListRules/index.jsx
```diff
@@ -557,17 +714,18 @@ function ListRules(props) {
...
+        onExited={handleDialogExited}
```
helfi92: Setting `open: false` sounds like the right thing to do. I wonder if we actually need the `onExited` function. The UI behavior seems to be unchanged whether `handleDialogExited` is called or not. There's also no problem leaving it here. Your call.

src/components/RuleCard/index.jsx
```diff
@@ -751,6 +772,8 @@ RuleCard.propTypes = {
   onRuleDelete: func,
   // If true, the card will hide all buttons.
   readOnly: bool,
+  onSignoff: func.isRequired,
+  onRevoke: func.isRequired,
```
helfi92: `onSignoff` and `onRevoke` should not be required when `readOnly` is set to true. I would remove the `isRequired` and add a comment saying that those properties are required if `readOnly` is false.

mozbhearsum: Good point!

src/views/Rules/ListRules/index.jsx
mozbhearsum: Just for cleanliness, I think it's probably best to clean up the state, so I'll keep it in.

PR #117, opened 2019-07-25
Adjust row height for rules

helfi92:
Does the spacing look similar for you?

Screen Shot 2019-07-25 at 4 28 37 PM

### DIFF
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -218,430 +218,433 @@ function RuleCard({ rule, onRuleDelete, readOnly, ...props }) {
         />
       )}
       <CardContent classes={{ root: classes.cardContentRoot }}>
-        <Grid container>
-          <Grid item xs={4}>
-            <List>
-              {rule.mapping && (
-                <ListItem className={classes.listItem}>
-                  <ListItemText
-                    title={rule.mapping}
-                    primaryTypographyProps={{
-                      component: 'div',
-                      className: classes.primaryText,
-                    }}
-                    secondaryTypographyProps={{
-                      className: classes.textEllipsis,
-                    }}
-                    primary={
-                      <Fragment>
-                        Mapping
-                        {diffedProperties.includes('mapping') &&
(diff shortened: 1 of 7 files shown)


src/components/RuleCard/index.jsx
```diff
@@ -218,430 +218,433 @@ function RuleCard({ rule, onRuleDelete, readOnly, ...props }) {
...
-                    secondary={rule.headerArchitecture}
-                  />
-                </ListItem>
-              )}
-            </List>
-          </Grid>
-          <Grid item xs={12}>
-            {rule.comment && (
+        {(!rule.scheduledChange ||
+          rule.scheduledChange.change_type !== 'insert') && (
```
helfi92: The only difference in this file is that the <Grid  container /> element is now wrapped in a condition so that we don't render an empty grid container when the grid items are hidden.

helfi92:
Forgot to change something. I'll flag for review when ready.

mozbhearsum:
I still have a few smooshed cards:
image

Which I _think_ are all caused by multiline comments in the diff view (and I'm happy to consider that a different bug).

Other than that, it looks great to me.

mozbhearsum:
I just dug into the smooshing a bit more and it looks like it's partly caused by the multiline comment. Here's what it looks like with a single line:
image

But it looks like there's something else going on there too. Again, happy to consider it a different bug if it makes sense to.

helfi92:
@mozbhearsum To avoid multilines, I added a patch to remove word wrapping and use a scroller instead. Do you mind having another look, please? The tradeoff with this is that the red and green boxes can have varying widths.

helfi92:
@mozbhearsum I pushed a patch. Do you mind taking a look again and perhaps sending a screenshot, please? Thanks!

mozbhearsum:
This looks good to me:
image

The extra gap in the second rule is caused by the display issue for jaws/mig64 that I'm fixing in #101, so I wouldn't worry about it.

Issue #98, opened 2019-07-19
jaws & mig64 values change unexpectedly


mozbhearsum:
While testing something unrelated, I noticed that jaws and mig64 tried to change from null -> false when I only meant to change another field:
![image](https://user-images.githubusercontent.com/49649/61560844-70933a00-aa3b-11e9-9357-bae7dfb7f233.png)

It looks like the frontend translates false into null at some point.

helfi92:
Good catch. We should fix that.

helfi92:
I believe this was fixed.

PR #118, opened 2019-07-25
Interpret jaws/mig64 to the same values that the UI sets them to

mozbhearsum:
The root of this issue was that these fields come in as null/false/true, but the UI expects them (and sets them to, when the value changes) '-'/'false'/'true'. I tried to change things in a way where it would set them to the same thing as the backend, but React doesn't like values being set to null, and boolean always get cast to strings AFAICT -- so this seemed like the next best solution. Alternatively, we could support both types of values in the handlers/display code, but that's a bit more kludgy and error prone IMO.

### DIFF
--- a/src/views/Rules/Rule/index.jsx
+++ b/src/views/Rules/Rule/index.jsx
@@ -253,6 +253,14 @@ export default function Rule({ isNewRule, ...props }) {
   };
 
   useEffect(() => {
+    // Some fields (jaws and mig64 at the time of writing) are optional
+    // boolean fields. Because input fields can't handle null or boolean
+    // values, they get stored as EMPTY_MENU_ITEM_CHAR or a string
+    // true/false when they change in the UI. In order to keep things
+    // consistent, we do the same to the data fetched from the server.
+    const getOptionalBooleanValue = current =>
+      current === null ? EMPTY_MENU_ITEM_CHAR : String(current);
+
     if (!isNewRule) {
       Promise.all([
         fetchRule(ruleId),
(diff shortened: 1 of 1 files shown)


helfi92:
I'm still seeing `undefined` for `jaws` as well as the other properties when inserting a rule:

Screen Shot 2019-07-26 at 1 00 59 PM

What do you think about

```diff
diff --git a/src/utils/diff.js b/src/utils/diff.js
index 6344535..4d86700 100644
--- a/src/utils/diff.js
+++ b/src/utils/diff.js
@@ -16,7 +16,7 @@ export default (diffProperties, objectOne, objectTwo) => {
       }
 
       return {
-        prev: `${title(prop)}: ${prev}`,
+        prev: `${title(prop)}: ${prev === undefined ? null : prev}`,
         next: `${title(prop)}: ${next}`,
       };
     })
```

mozbhearsum:
Oh sorry. While I was working on investigating #98 I discovered another bug which caused us to be setting jaws and mig64 to the wrong values. That's what I'm fixing here (I just adjusted the title again to stop claiming I'm fixing #98...I clearly got myself confused, too).

The specific issue is that the jaws and mig64 fields get set to improper values when saying a Rule in some cases. For example, if you click "Update" on a Rule that currently has them both set to null, then click "Save" without making any modifications, they'll both get set to false. This is because of https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/views/Rules/Rule/index.jsx#L212, where we're expecting EMPTY_MENU_ITEM_CHAR for null values.

More generally, when we load the data from the backend, jaws and mig64 are either null, true, or false. When we _change_ them with the dropdown, they end up as '-', 'true', or 'false'. The mutation logic handles this fine already, so this patch adjusts the data loaded from the backend to match those values.

(I originally tried to do the opposite and use null/true/false for the changed values, but React complained about a value being set to null.)

mozbhearsum:
> Looks good. Thank you for the explanation. That made things a lot clearer.
> 
> > (I originally tried to do the opposite and use null/true/false for the changed values, but React complained about a value being set to null.)
> 
> String(null), String(true), String(false) may be helpful. The text fields should accept null as a string.

Good to know. I think this would still mean I need to cast them back to bool or null in the event handlers though? In which case it doesn't make much of a difference (just checking for 'null' instead of EMPTY_MENU_ITEM_CHAR as best I can tell).

PR #120, opened 2019-07-29
Releases adjust height

helfi92:


src/components/ReleaseCard/index.jsx
```diff
@@ -106,7 +106,7 @@ function ReleaseCard(props) {
         className={classes.cardHeader}
         classes={{ content: classes.cardHeaderContent }}
         title={
-          <Typography component="h2" variant="h6">
+          <Typography className={classes.ruleName} component="h2" variant="h6">
```
mozbhearsum: Nit: I think this should be called `releaseName` instead.

PR #123, opened 2019-07-29
Enhance the AutoCompleteText component

helfi92:
Features:
* Add arrow to show it's a dropdown
* Allow users to click the down keyboard button to see the list of options
* Make chips smaller

### DIFF
--- a/src/components/AutoCompleteText/getSuggestions.js
+++ b/src/components/AutoCompleteText/getSuggestions.js
@@ -1,19 +1,11 @@
 export default suggestions => value => {
   const inputValue = value.trim().toLowerCase();
   const inputLength = inputValue.length;
-  let count = 0;
 
   return inputLength === 0
-    ? []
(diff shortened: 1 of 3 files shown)


src/components/AutoCompleteText/getSuggestions.js
```diff
@@ -1,19 +1,11 @@
...
-          count += 1;
-        }
-
-        return keep;
-      });
+    ? suggestions
+    : suggestions.filter(
+        suggestion =>
+          suggestion.slice(0, inputLength).toLowerCase() === inputValue
+      );
```
mozbhearsum: Can we sort alphabetically here, while we're at it? Or is that better done in the components that use this?

helfi92: I think it's best to do it in the component that uses it so that we can use `useMemo` to memoize the list and avoid having to sort the list on each render. I'll file an issue for that.

PR #125, opened 2019-07-29
Skeleton UI for scheduled user permission changes (fixes #90)

### LINKED ISSUE #90: display required & pending signoffs on user permission scheduled changes

mozbhearsum:
I largely took inspiration from the Required Signoffs UI for this one, using the arrows to highlight changes instead of a diff viewer. The best idea I could come up with for showing these is one additional section per pending change. If there's many queued up this will obviously get rather lengthy, but permission scheduled changes are relatively rare and short lived, so this is unlikely to be the case.

There's a couple of rough edges that I'm not 100% happy with as well. They're not dealbreakers, but if you have any ideas I'd appreciate them:
* Pending inserts and pending updates look very similar at a glance. They both use the same `PENDING` badge; the only differences are the lack of red dot on the original permission, and the `->` in the pending change.
* On longer permission descriptions, the text for the scheduled change version wraps all the way back to the current permission, which I find a little confusing - it can look like it's the text from the current permission wrapping instead. I couldn't figure out how to fix this.

image

This will need rebasing after #101 lands.

### DIFF
--- a/src/components/UserCard/index.jsx
+++ b/src/components/UserCard/index.jsx
@@ -100,41 +100,45 @@ function User(props) {
           />
         </CardActionArea>
       </Link>
-      <CardContent>
-        <List>
-          {Object.entries(permissions).map(([permission, details]) => (
-            <ListItem key={permission}>
-              <ListItemIcon>
-                <KeyVariantIcon />
-              </ListItemIcon>
-              <ListItemText
-                primary={
-                  <Fragment>
-                    {getPermissionString(
-                      permission,
-                      returnOptionIfExists(details.options, 'actions', []),
-                      returnOptionIfExists(details.options, 'products', [])
-                    )}
-                    {scheduledPermissions[permission] && (
-                      <span className={classes.propertyWithScheduledChange} />
-                    )}
-                  </Fragment>
-                }
-              />
-            </ListItem>
-          ))}
-          {Object.keys(roles).length > 0 && (
-            <ListItem>
-              <ListItemIcon>
-                <AccountGroupIcon />
-              </ListItemIcon>
-              <ListItemText>
-                holds the {getRolesString(Object.keys(roles))}
-              </ListItemText>
-            </ListItem>
-          )}
-        </List>
-      </CardContent>
+      {/* We don't need to check roles here because users without permissions
+            cannot hold roles */}
+      {Object.keys(permissions).length > 0 && (
+        <CardContent>
+          <List>
+            {Object.entries(permissions).map(([permission, details]) => (
+              <ListItem key={permission}>
+                <ListItemIcon>
+                  <KeyVariantIcon />
+                </ListItemIcon>
+                <ListItemText
+                  primary={
+                    <Fragment>
+                      {getPermissionString(
+                        permission,
+                        returnOptionIfExists(details.options, 'actions', []),
+                        returnOptionIfExists(details.options, 'products', [])
+                      )}
+                      {scheduledPermissions[permission] && (
+                        <span className={classes.propertyWithScheduledChange} />
+                      )}
+                    </Fragment>
+                  }
+                />
+              </ListItem>
+            ))}
+            {Object.keys(roles).length > 0 && (
+              <ListItem>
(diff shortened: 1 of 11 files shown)


mozbhearsum:
I like these changes, I'll pull them in - thank you!

mozbhearsum:
OK, this is pretty much the same as your screenshot now. I think I did something slightly different with the new `<Grid>` however, as I'm not able to use xs=5, 1, and 6 without it wrapping the entire last `<Grid>`.

src/components/UserCard/index.jsx
```diff
@@ -59,34 +110,123 @@ export default function User(props) {
...
+                          []
+                        ),
+                        returnOptionIfExists(
+                          permissions[permission].options,
+                          'products',
+                          []
+                        )
+                      )}
+                    </Grid>
+                    <Grid item xs={1}>
```
helfi92: Center the arrow in the grid. I think you could do `display: 'flex', justifyContent: 'center'` on the Grid.

src/components/UserCard/index.jsx
```diff
@@ -59,34 +110,123 @@ export default function User(props) {
...
+                      <Fragment>
+                        <ArrowRightIcon className={classes.arrowIcon} />
+                      </Fragment>
```
helfi92: We probably don't need `Fragment`.

src/components/UserCard/index.jsx
```diff
@@ -26,17 +35,61 @@ const useStyles = makeStyles(theme => ({
...
   },
+  scheduledChangeDescriptionMarginLeft: {
+    lineHeight: '24px',
+    marginLeft: theme.spacing(1),
+    marginBottom: theme.spacing(2),
+  },
+  scheduledChangeDescription: {
+    lineHeight: '24px',
+    marginBottom: theme.spacing(2),
+  },
```
helfi92: Instead of having `marginBottom` on both `scheduledChangeDescriptionMarginLeft` and `scheduledChangeDescription`, we could define it once on the Grid container.

src/components/UserCard/index.jsx
```diff
@@ -26,17 +35,61 @@ const useStyles = makeStyles(theme => ({
...
+  propertyWithScheduledChange: {
+    width: theme.spacing(1),
+    height: theme.spacing(1),
+    borderRadius: '50%',
+    background: 'red',
+    marginLeft: theme.spacing(1),
+    display: 'inline-block',
+  },
```
helfi92: We should probably expose this red dot in `theme.js` under mixins.


```js
// theme.js
redDot: {
  // ...
},
```

```jsx
// UserCard/index.jsx
propertyWithScheduledChange: {
  ...theme.mixins.redDot,
},

// RuleCard/index.jsx
propertyWithScheduledChange: {
  ...theme.mixins.redDot,
},
```

src/components/UserCard/index.jsx
```diff
@@ -95,9 +235,15 @@ User.propTypes = {
...
+  onRevoke: func.isRequired,
 };
 
 User.defaultProps = {
   roles: {},
   permissions: {},
+  scheduledPermissions: {},
 };
+
+export default withUser(User);
```
helfi92: I think adding `marginBottom: theme.spacing(4)` to the Grid container will make it easier to distinguish the change from the required signoffs:

Screen Shot 2019-07-30 at 9 42 02 AM

helfi92:
I pushed minor UI changes to this branch. The UI now looks like:

Screen Shot 2019-07-31 at 11 00 21 AM

Main changes:
* I wasn't satisfied with the location of the arrow icon. I removed the `Grid` and used a plain flex `div`
* Changed font style for the permission changes to make the change stand out more
* Modifed the required signoff view a bit to make sure they are both consistent
* Left aligned everything in the user card with the title in order to be consistent with the required signoff card


Hope that's okay. Let me know if you find anything weird or something doesn't make sense.

mozbhearsum:
This looks great, I especially like the altered font style. Thanks!

PR #126, opened 2019-07-30
Implement add user functionality

mozbhearsum:
`ViewUser` more or less supports this already, except I hid the Roles section for new users, because they cannot be added until a User has a permission (and not just a scheduled one).


src/views/Users/ViewUser/index.jsx
```diff
@@ -445,22 +445,26 @@ function ViewUser({ isNewUser, ...props }) {
...
+                      className={classes.fullWidth}
+                      variant="outlined">
+                      <PlusIcon />
+                    </Button>
+                  </Grid>
+                </Grid>
+                <br />
+                <br />
+                <br />
+              </Fragment>
```
helfi92: When the browser is resized in such a way that a user needs to scroll to reach the end of the page, the plus button has no bottom margin:

Screen Shot 2019-07-31 at 8 39 45 AM

This should fix it:

```diff
diff --git a/src/views/Users/ViewUser/index.jsx b/src/views/Users/ViewUser/index.jsx
index ea2521d..5606082 100644
--- a/src/views/Users/ViewUser/index.jsx
+++ b/src/views/Users/ViewUser/index.jsx
@@ -41,6 +41,7 @@ const useStyles = makeStyles(theme => ({
   },
   addGrid: {
     marginTop: theme.spacing(5),
+    marginBottom: theme.spacing(10),
   },
   gridWithIcon: {
     marginTop: theme.spacing(3),
@@ -460,9 +461,6 @@ function ViewUser({ isNewUser, ...props }) {
                     </Button>
                   </Grid>
                 </Grid>
-                <br />
-                <br />
-                <br />
               </Fragment>
             )}
             <Typography variant="h5">Permissions</Typography>
```

src/views/Users/ViewUser/index.jsx
```diff
@@ -445,22 +445,26 @@ function ViewUser({ isNewUser, ...props }) {
...
+            )}
             <Typography variant="h5">Permissions</Typography>
```
helfi92: When adding a new user, the product restrictions autocomplete text fields has no suggestions:

Screen Shot 2019-07-31 at 8 41 22 AM 

Is this expected?

mozbhearsum: No, hm. I'll look into this.

src/views/Users/ViewUser/index.jsx
```diff
@@ -115,79 +116,84 @@ function ViewUser({ isNewUser, ...props }) {
...
+      }
+
+      setRoles(roles);
+      setOriginalRoles(clone(roles));
+      setPermissions(permissions);
+      setOriginalPermissions(clone(permissions));
+      setProducts(productdata.data.data.product);
+      setRequiredSignoffs(rs.data.data.required_signoffs);
+    });
   }, []);
```
helfi92: Navigating to http://localhost:9000/users/balrogadmin then selecting a product restriction for the last permission with the name "rule" throws an error in the console `TypeError: result.sc.options is null` and then shows a blank page. Works for the product restriction of "release_read_only" however.

mozbhearsum: Is that permission currently a scheduled change, or a full fledged permission? There's probably something busted in the `useEffect`, but I'm having trouble reproducing.

helfi92: Screen Shot 2019-07-31 at 4 15 53 PM

mozbhearsum: Hm, I think after my recent changes it's no longer possible to get into this state. However, I'm going to push a fix that will ensure that even if sc.options from the server is null, it doesn't end up that way in the state.

src/views/Users/ViewUser/index.jsx
```diff
@@ -164,6 +164,10 @@ function ViewUser({ isNewUser, ...props }) {
                 // Array destructuring is not very readable here.
                 // eslint-disable-next-line prefer-destructuring
                 permission.sc = sc[0];
+
+                if (permission.sc.options === null) {
```
helfi92: Could this be `undefined`? I wonder if we should just do `if (!permission.sc.options)` which would catch both `undefined` and `null`.

src/views/Users/ViewUser/index.jsx
```diff
@@ -117,7 +121,9 @@ function ViewUser({ isNewUser, ...props }) {
 
   useEffect(() => {
     Promise.all([
-      isNewUser ? null : fetchUser(existingUsername),
+      userExists(existingUsername).then(exists =>
+        exists ? fetchUser(existingUsername) : null
+      ),
```
helfi92: Instead of using `userExists`, any chance we can use `withUser`? I think we would be able to rewrite to something like:

```jsx
Promise.all([
  user ? fetchUser(existingUsername) : Promise.resolve(),
  // ...
]);
```

mozbhearsum: Unless I'm misunderstanding, I don't think that will work here, because the username in question isn't the logged in user, it's the user we're creating or editing.

src/views/Users/ViewUser/index.jsx
mozbhearsum: Yeah, that's a good idea.

src/views/Users/ViewUser/index.jsx
helfi92: I think fetching a user that doesn't exists should show an error to the user as a form of action feedback.

Also, the issue I was having was on the view that lists all users, not a single user. I don't seem to get this error anymore for some reason.

mozbhearsum: Okay. I still think we need to do something here though - there was an error that did show up on ViewUser for a user that only has scheduled permissions.

helfi92: > there was an error that did show up on ViewUser for a user that only has scheduled permissions.

Would it make sense to have the backend not throw and error but instead return an empty set when performing the request that is currently throwing the error?

mozbhearsum: Thinking from the backend perspective I don't think this makes sense - trying to fetch a user that doesn't exist seems like a legitimate 404 to me. I'll see if I can come up with another way to address this.

helfi92: I agree with you. Maybe we can just catch the error. We could probably use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled when FF supports it but for the time being one approach I think we do is take the asynchronous call outside the `Promise.all` so that it doesn't stop the other calls for proceeding.

mozbhearsum: > Maybe we can just catch the error

I was looking into this angle, and our interceptor already catches and throws a new Error: https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/App.jsx#L32. We could catch that, of course, but without the actual axios response we have to check the error message itself to know whether it was a 404 or something else, which is pretty nasty. This makes me wonder if the interceptor needs to change...

src/views/Users/ListUsers/index.jsx
```diff
@@ -58,15 +57,16 @@ function ListUsers() {
...
+          <Link to="/users/create">
+            <Tooltip title="Add User">
+              <Fab
+                color="primary"
+                className={classes.fab}
+                classes={{ root: classes.fab }}>
+                <PlusIcon />
+              </Fab>
+            </Tooltip>
+          </Link>
```
helfi92: Navigating to http://localhost:9000/users/create throws an error `User does not exist!`.

mozbhearsum: D'oh, silly mistake.

helfi92:
How does test the creation of a user? I'm not sure I understand these error messages.

Screen Shot 2019-08-07 at 11 30 28 AM

Screen Shot 2019-08-07 at 11 32 17 AM

mozbhearsum:
Okay, so the first one is actually a bug in the backend form validation. I'm going to open a PR to fix that shortly. The second one is a bug on this side. I'll get that fixed and do some additional testing before the next round of review.

mozbhearsum:
I hesitate to mark this ready for review again, but I've fixed the issues noted above, and a couple of other small things I found in testing. The issue with the `permission` permission (heh) was caused by incorrect data in `constants.js` and a missing check in `updateUser.js`. The other fixes are:
- Don't show SignoffSummary nor actions if a change doesn't require signoff.
- Sort users before displaying them.

helfi92:
Awesome. I will review soon :)

src/views/Users/ViewUser/index.jsx
```diff
@@ -115,79 +122,98 @@ function ViewUser({ isNewUser, ...props }) {
...
       });
-    }
+
+      if (!isNewUser && permissions.length === 0) {
+        setUserError('User does not exist!');
+
+        return;
+      }
+
+      setUsername(existingUsername);
```
helfi92: It would be nice to have `setUsername` come before L203 so that when a user tries to view a username that doesn't exist, the UI will have the username field prefilled.

_Example: Navigating to http://localhost:9000/users/abcd then clicking the "X" button of the error dialog_

Screen Shot 2019-08-07 at 8 25 11 PM

mozbhearsum: Ah, that's a good point. Will do.

mozbhearsum: Is there any reason we can't set it at line 89 as the default value in `useState`? It's available there, so maybe that's the best spot, since it would ensure it's set regardless of any errors (including networking errors fetching data).

helfi92: I like that.

src/views/Users/ViewUser/index.jsx
```diff
@@ -86,7 +86,7 @@ function ViewUser({ isNewUser, ...props }) {
     },
   } = props;
   const classes = useStyles();
-  const [username, setUsername] = useState('');
+  const [username, setUsername] = useState(existingUsername);
```
helfi92: Thoughts on making this `const [username, setUsername] = useState(isNewUser ? '' : existingUsername);`?

mozbhearsum: Probably a good idea, for some reason I thought existingUsername would be `''` in the new user case...

PR #131, opened 2019-08-01
Move CSS import to DiffRule

helfi92:
Not sure why I didn't move this earlier but this will make sure other views will be able to use the diff viewer without having to load the rules view first.


helfi92:
> This doesn't fix the issue AFAICT. I applied it, built, and ran the app through my local apache install and I'm still missing the same CSS. I tried moving it to ListRules as well, which also didn't help.

This won't necessary solve the issue you are seeing in staging but still needs to be done. Any chance some styles are being blocked by CSP?

mozbhearsum:
> This won't necessary solve the issue you are seeing in staging but still needs to be done. Any chance some styles are being blocked by CSP in staging?

Not as far as I can tell. There's no CSP yet, and no errors in the Console.

PR #135, opened 2019-08-02
Proof of concept implementation of rewind-the-rules-table UI.

bhearsum:
This is not remotely complete or ready for review, but it's a vaguely working proof of concept for showing the Rules table at a point in time. It requires the backend to be running https://github.com/mozilla/balrog/compare/master...mozbhearsum:point-in-time?expand=1


nthomas-mozilla:
I had a brief experiment with these changes, looks good. It might be useful try modifying the banner so it reads `Balrog Admin ┃ Rules @ <date and time>`, since that'll remain on screen while the picker will very likely have been scrolled off to reach the active rules. Not sure if you are finished with changing the Duplicate/Update/Delete buttons on rules, but they seem to still be working links even though they are grey now.

mozbhearsum:
> It might be useful try modifying the banner so it reads Balrog Admin ┃ Rules @ <date and time>

I like this, let's do it!

Thanks for the report about the buttons, I'll get that fixed, too.

mozbhearsum:
This is still very rough, but I managed to clean things up a little bit, and implement a basic diff mode:
image

I'm using a terrible hack of faking a scheduled change to make the diff mode work, which obviously has to be cleaned up if we move forward here. The performance is also not great, for reasons I don't fully understand (I'm guessing that some of my hacks are causing them, but I haven't dug deep into that).

Before I spend the time on that, could I get some feedback to see if there's any general changes that need to be made, first?

mozbhearsum:
Looks like I can't formally request anything from @nthomas-mozilla, so I'll just mention you here instead :).

src/components/RuleCard/index.jsx
```diff
@@ -771,6 +778,8 @@ RuleCard.propTypes = {
   onRuleDelete: func,
   // If true, the card will hide all buttons.
   readOnly: bool,
+  // If true, the card will disable all buttons
```
helfi92: I think the revisions button in the rule card doesn't currently get disabled when `disableActions` is `true`. If this is intended, then we should probably mention it in the comment.

src/views/Rules/ListRules/index.jsx
```diff
@@ -630,11 +676,13 @@ function ListRules(props) {
             : Object.values(rule.scheduledChange).join('-')
         }
         style={style}>
+        {/* should we go read only mode if rewindDAte is set instead? */}
         <RuleCard
           className={classes.ruleCard}
           key={rule.rule_id}
           rule={rule}
           onRuleDelete={handleRuleDelete}
+          disableActions={!!rewindDate}
```
helfi92: `Boolean(rewindDate)`.

src/views/Rules/ListRules/index.jsx
```diff
@@ -677,6 +725,24 @@ function ListRules(props) {
...
           <div className={classes.options}>
+            <DateTimePicker
+              disableFuture
+              inputVariant="outlined"
+              label="Rewind to..."
+              onError={handleRewindDateTimePickerError}
+              helperText={rewindDateError}
+              onDateTimeChange={handleRewindDateTimeChange}
+              value={rewindDate}
+            />
```
helfi92: On page load, the console throws ```Failed prop type: The prop `value` is marked as required in `DateTimePicker`, but its value is `null`.```.

src/views/Rules/ListRules/index.jsx
```diff
@@ -254,7 +259,22 @@ function ListRules(props) {
...
 
       setRulesWithScheduledChanges(sortedRules);
-    });
+      setRewoundRules([]);
+    }
+  }, [rules, scheduledChanges, requiredSignoffs]);
+
+  useEffect(() => {
+    fetchRules(rewindDate ? rewindDate.getTime() : null);
+  }, [rewindDate]);
```
helfi92: Do we `fetchRules` each time the `rewindDate` value change (i.e., a user clicking on different dates inside the widget)? This might explain why the picker is lagging.

helfi92 (commented):
Overall, this feature is really nice and I'm sure this will make other folks really happy :) We should definitely do it.

I wonder however if we should spend more time on this after the [MVP issues](https://github.com/mozilla-frontend-infra/balrog-ui/issues?q=is%3Aopen+is%3Aissue+label%3AMVP) are closed. Having an MVP balrog UI sometime during this quarter would be ideal :)

mozbhearsum:
> I wonder however if we should spend more time on this after the MVP issues are closed. Having an MVP balrog UI sometime during this quarter would be ideal :)

This is a fair point. I primarily did this as a break when I got stuck on other issues, and to make sure it was viable. I agree that it's less important than the MVP issues, so I'll be coming back to it after we fix them.

Issue #16, opened 2019-05-31
Refresh auth0 tokens in the background


mozbhearsum:
I thought this was already implemented, but I noticed while working on some write operations that I started getting errors about the token being expired (from the Balrog backend) after awhile, and had to re-login to fix them.

helfi92:
This is implemented in https://github.com/mozilla-frontend-infra/react-auth0-components/blob/99c32611abd9bc5ada5ca5daa8a944d08fa055d3/src/components/Authorize/index.jsx#L262-L296. We would need to dig inside and see why it's not working.

mozbhearsum:
This might be an issue with my auth0 configuration. I just added 'http://localhost:9000' to the Allowed Web Origins, and we'll see if that clears it up.

mozbhearsum:
I was able to reproduce this in stage today. STR:
* Open https://balrog-admin-static-stage.stage.mozaws.net/users
* Verify that you can make some sort of change (eg: edit a rule)
* Leave it open for more than 2 hours to let the token expire
* Try to make another change. You should get an error about no authorization token being passed.

As best I can tell, the refresh attempt never happens - I had the developer console open the entire time, and I saw no XHR requests (although I'm not sure if I would see XHRs that happen in an iframe, if that's how it attempts it).

When I inspected localStorage, it was empty - so my session was removed at some point too.

mozbhearsum:
I think this is reproducible the same way in local dev too (although I haven't tried yet), and I've got access to the auth0 logs for that if it helps.

helfi92:
Sounds good. I will try to investigate soon.

PR #136, opened 2019-08-02
Security headers for local development

mozbhearsum:
This is largely the same as what we'll use in stage/prod, except where noted to be different. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1571095 for applying the appropriate parts to stage.


mozbhearsum:
I just dug into unsafe-inline a bit more and it turns out my React Developer Tools addon was causing it to be required, not the app itself. I also discovered we need frame-src for token refreshes (although I don't think the lack of it was causing #16

.neutrinorc.js
```diff
@@ -17,6 +17,29 @@ module.exports = {
...
+                       (which can't be locked down to a specific path due to CSP format)
+              style-src: https://fonts.googleapis.com for dynamically loaded fonts
+              font-src: https://fonts.gstatic.com for fonts!
+              connect-src: https://localhost:9000 for the backend api
+                           {ws,https}://localhost:8010 for webpack connections (local dev only)
+                           https://balrog-localdev.auth0.com for authentication.
+                           note: this is different in stage/prod
+              frame-src: https://balrog-localdev.auth0.com for background token refreshes
+            */
+            'Content-Security-Policy': "default-src 'none'; script-src 'self' 'unsafe-eval'; img-src 'self' https://*.gravatar.com https://*.githubusercontent.com https://i1.wp.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; font-src https://fonts.gstatic.com; connect-src http://localhost:9000 ws://localhost:9000 https://localhost:8010 https://balrog-localdev.auth0.com; frame-src: https://balrog-localdev.auth0.com",
```
helfi92: For places where we have `localhost:9000`, I think we could just use `self` instead.

mozbhearsum: Good point. I'll do some double checking to make sure this works with ws://, too.

mozbhearsum:
Whoops, sorry. It seems I didn't retest after adding frame-src (which I bungled). Should be fixed now.

mozbhearsum:
> Ah, we would also probably want to allow the endpoint for release revisions coming in from googleapis.

Good point, I'll add that.

Issue #64, opened 2019-07-08
UI for releases history


mozbhearsum:
A view that shows changes to a Release over time should be enough here. Being able to diff between Release revisions is important for understandability.

Unlike other parts of Balrog, Release history is stored in GCS, and it must be queried for directly there (not through the backend). This is done through queries like, eg: https://www.googleapis.com/storage/v1/b/balrog-prod-nightly-history-v1/o?prefix=Firefox-mozilla-central-nightly-latest/&delimeter=/.

helfi92:
I can try to work on one of the history views next. I will be a bit busy with Taskcluster this week so progress will be slow but fwiw, that's what I'm planning to tackle next. Let me know if I should focus my energy elsewhere.

helfi92:
I noticed that the Product/Channel filter of a rule revision is always empty. Is this always the case?

_Example: https://localhost:8010/rules/306_

<img width="1200" alt="Screen Shot 2019-07-18 at 6 54 36 PM" src="https://user-images.githubusercontent.com/3766511/61497473-834f3580-a98d-11e9-9a7f-2319424635c6.png">

mozbhearsum:
It doesn't really make sense to show it at all in the context of a single Rules history. Looks like a bug to me!

helfi92:
We could probably use a similar UI to the rule revisions view.

helfi92:
I'm currently trying to think of a possible layout for this view. The view in the old site looks like

![Screen Shot 2019-08-06 at 12 32 55 PM](https://user-images.githubusercontent.com/3766511/62558163-7adc7300-b846-11e9-8d7c-7a0a9d5683ef.png)


It would probably be good to make the layout similar to the roles revisions view. 2 questions come to mind:

1. Does "View Diff" show the diff with the current release? Would it be good if we use a similar layout to the rules revisions view (https://github.com/mozilla-frontend-infra/balrog-ui/pull/102#issue-299921309?

2. "View Data" shows the JSON representation of the release. Would it make sense to show a release card instead via the `ReleaseCard` component. We would probably need to introduce a `readOnly` prop similar to `RuleCard` to make sure action buttons are hidden.

mozbhearsum:
> Does "View Diff" show the diff with the current release? Would it be good if we use a similar layout to the rules revisions view (#102 (comment)?

[...]

Issue #104, opened 2019-07-22
Virtualize list of rule and release revisions


helfi92:
Reference: https://github.com/mozilla-frontend-infra/balrog-ui/pull/102#issuecomment-513885645

Solution: Use https://github.com/bvaughn/react-virtualized

helfi92:
@mozbhearsum I was wondering if we ever need more than the last 50 revisions of a rule or release. Virtualization may not be needed after all.

mozbhearsum:
For Rules it's certainly rare (but occasionally needed). For releases it's much more common to need to dig back hundreds, or even more than a thousand revisions back.

helfi92:
Gotcha. Let's virtualize :)

helfi92:
We should be able to copy-paste some of the logic in https://github.com/mozilla-frontend-infra/balrog-ui/pull/137 to virtualize the rules list. I can do that once #137 is merged.

PR #137, opened 2019-08-06
Add UI for release revisions

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/64.
Relates to #104.

Note: I edited the title of https://github.com/mozilla-frontend-infra/balrog-ui/issues/104 to also virtualize the list of release revisions.


helfi92:
Screen Shot 2019-08-07 at 11 25 09 AM

src/App.jsx
```diff
@@ -10,6 +10,10 @@ import theme from './theme';
 import Main from './Main';
 
 axios.interceptors.request.use(config => {
+  if (config.url.startsWith('http')) {
```
mozbhearsum: I assume this is to avoid sending credentials to Google and/or Auth0. Would it be better to check for a specific domain(s) instead of a protocol? Looking for just `http` feels like it's bound to break in the future. Or perhaps this is an argument for using Axios instances instead of a global? The latter might be out of scope for this PR.

src/services/releases.js
```diff
@@ -1,10 +1,66 @@
...
+  }
+
+  async function getReleases(url, pageToken) {
+    const response = await axios.get(
+      pageToken ? `${url}&pageToken=${pageToken}` : url
+    );
+
+    parseReleases(response.data.items);
+
+    if (response.nextPageToken) {
```
mozbhearsum: I think this needs to be `response.data`.

src/utils/constants.js
```diff
@@ -5,6 +5,12 @@ export const OBJECT_NAMES = {
   PERMISSIONS_REQUIRED_SIGNOFF: 'required_signoffs/permissions',
   PRODUCT_REQUIRED_SIGNOFF: 'required_signoffs/product',
 };
+export const GCS_CONFIG = {
+  NIGHTLY_HISTORY_BUCKET:
+    'https://www.googleapis.com/storage/v1/b/balrog-prod-nightly-history-v1/o',
```
mozbhearsum: This is going to differ depending on the environment we're deployed in. Can we read it from the environment? These are the correct URLs for local dev, so they can go in `.neutrinorc.js`.

src/utils/diff.js
```diff
@@ -17,12 +33,8 @@ export default (diffProperties, objectOne, objectTwo) => {
...
 
       return {
-        prev: `${title(prop)}: ${
-          typeof prev === 'string' ? prev.replace(NEW_LINES_REGEX, '') : prev
-        }`,
-        next: `${title(prop)}: ${
-          typeof next === 'string' ? next.replace(NEW_LINES_REGEX, '') : next
-        }`,
+        prev: `${title(prop)}: ${formatValue(prev)}`,
+        next: `${title(prop)}: ${formatValue(next)}`,
```
mozbhearsum: If possible, can we avoid title casing propreties within the release data? It's jarring to have them differ between the full release data view and the diff (and the users who edit the blobs are accustomed to the specific property names).

src/services/releases.js
```diff
@@ -1,10 +1,66 @@
...
+    }
+
+    // descending sort, so newer versions appear first
+    return releases.sort((a, b) => a.data_version < b.data_version);
+  }
+
+  return getReleases(`${bucket}?prefix=${name}/&delimeter=/`);
+};
+
+const getRevisionData = link => axios.get(link);
```
mozbhearsum: Should the existing consumers of this just call `axios.get` instead, or is it bad form/architecture to do that directly in the views/components? I don't have super strong feelings here, it just stands out to see a function that does nothing except pass the arguments to another function.

helfi92:
> The equivalent on the old UI (https://localhost:8010/releases/Devedition-58.0b10-build1) takes ~2s to load and render on my laptop.

> I also can't get the page to work correctly for releases with a lot of history like the one above - Firefox keeps throwing the slow script error until I decide to kill it.

I think this is because we are trying to display 1000 revisions using material UI components. The fix for that is to virtualize the list. This unacceptable performance hit is  the reason why I added the MVP label to https://github.com/mozilla-frontend-infra/balrog-ui/issues/104.

src/utils/diff.js
helfi92: Agreed. We should probably remove it alltogether even from the rules view.

src/App.jsx
helfi92: I think we never want to send the access token nor set `result.baseURL = BASE_URL;` when we are not sending a request to `BASE_URL`. I added a patch to add all of this logic inside a conditional:

```jsx
axios.interceptors.request.use(config => {
  const result = config;

  if (!config.url.startsWith('http')) {
    // set Authorization header
    // set baseURL
  }

  return result;
});
```

I think this will make it resilient to any future additions to that function. Let me know if you disagree. Happy to discuss.

helfi92:
i found a small UI glitch where the table header goes above the app bar when the table is big:

Screen Shot 2019-08-07 at 9 03 34 PM

This will most probably be fixed in https://github.com/mozilla-frontend-infra/balrog-ui/issues/104.

src/App.jsx
mozbhearsum: Ah, I think I get it now. Any Balrog backend URLs will only have a path component in them, whereas others will be a full URL.

helfi92: Yep.

helfi92:
Screen Shot 2019-08-09 at 11 19 08 AM


A couple of changes:

1. Table is now virtualized
2. I changed the diff display to something similar to `git diff`. The diff display is virtualized to make sure long JSON don't slow down the page or take forever to render.
3. Diffing big JSONs with thousand of lines can take a long time. I extracted the diffing logic into a web worker to make sure it doesn't slow down the UI which also allows us to show a spinner while the diffing logic is running.

helfi92:
> https://localhost:8010/releases/Devedition-56.0b1-build5

I'm not able to see that release for some reason even after searching for it in the releases view. All of the releases I'm seeing on that site seem to show only a couple of lines of diffs. Is it possible that we're diffing the wrong data in the new UI?

_Example: https://localhost:8010/releases/Devedition-56.0b3-build1_

Screen Shot 2019-08-09 at 1 56 18 PM

mozbhearsum:
> I'm not able to see that release for some reason even after searching for it in the releases view.

That's a little surprising...but the production data dump is updated every night, so we're probably just working with slightly different datasets.

> All of the releases I'm seeing on that site seem to show only a couple of lines of diffs.

Yep, this is pretty normal. Each revision is typically a single locale being added to the Release, so just the few lines between data versions is expected. If you diff, eg: version 1000 vs 500 you'll see ~500 locales being added.

> Is it possible that we're diffing the wrong data in the new UI?

Possibly...I'm going to dig a little bit more.

mozbhearsum:
After some print debugging, it looks to me like `diffLines` is never returning when run on large strings. I tried to do some debugging with a simple script, and it looks like it's stuck in GC and one of the diff.js methods mostly:
image

helfi92:
OK, I think I figured out why it's taking so long. What we need to do is deeply sort the objects first before comparing them.

helfi92:
Filed https://github.com/mozilla-frontend-infra/balrog-ui/issues/145 to track the performance issue when viewing a release.

mozbhearsum:
The diffs still work, but I'm getting this exception whenever I try change the versions being compared, usually multiple times (possibly once per hunk of the diff?):
`Warning: Encountered two children with the same key, `+        },`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`

I also saw an issue where there was weird looking hunks being shown, but I'm struggling to find a way to reproduce...

mozbhearsum:
The perf is much, much, much better! No more fan spinning forever :). A couple of other ideas for improvement (probably for later):
- Prefetch blobs onHover of the `View` button. Right now it takes long enough to be mildly annoying.
- Cache fetched blobs to improve successive diffs. We'd have to have a maximum size here to avoid blowing up memory usage, I suspect.

The other issue I saw is when I try comparing against a non-latest revision. For example, there's extra hunks at the top here:
image

It doesn't look like it happens with all releases, but I can always reproduce it with the CDM ones.

We can also see here that the revisions list is taking up more vertical space than necessary. Can we reduce that when there are few enough revisions that there's no scrolling?

This error is still present on first load of a revisions page too: `Warning: Failed prop type: Invalid prop `className` of type `string` supplied to `JsDiff`, expected `object`.`

helfi92:
Added the requested changes. I had to re-virtualize jsDiff in case there are >1k lines of diff. If you compare revisions that are far apart, I was able to find revisions with >5k of line diffing. We would probably need to have the diffing happen in a worker so that we can maybe show a loader or something and not have the main thread hang but I think I'll do that in a follow-up. I'll just have to copy-paste from one of the earlier commits.

helfi92:
Rebased.

src/components/Diff/index.jsx
```diff
@@ -1,9 +1,9 @@
 import React, { useMemo } from 'react';
+import { object, oneOf } from 'prop-types';
```
mozbhearsum: I see this file is getting renamed, but is still only used for diffing Rules. Similarly, `JsDiff` is only used for diffing Releases. I've got no objection to them being generic, but I think they could use better names. It's not clear to me what the difference between them is based on the current names.

PR #142, opened 2019-08-08
Schedule rule deletes when signoffs are required (fixes #115)

### LINKED ISSUE #115: delete button should schedule deletion when signoffs are required on rules page
Right now, it tries to directly delete, which fails.

mozbhearsum:
This works, but needs an extra fix to make `rowHeight` get recalculated. I dug into that part a bit, and it turns out that the `List` we use has a `recomputeRowHeights` function to do just that. I tried to get that working with a `ref` (as suggested in https://github.com/bvaughn/react-virtualized/issues/545), but our architecture doesn't seem to allow it - I get an error about needing to use a `ForwardRef` instead, which I wasn't able to get working.

Before I dive deeper down that rabbit hole, is that the right way to go, or is there a better solution here?

### DIFF
--- a/src/views/Rules/ListRules/index.jsx
+++ b/src/views/Rules/ListRules/index.jsx
@@ -303,7 +303,6 @@ function ListRules(props) {
           }),
     [productChannelFilter, rulesWithScheduledChanges]
   );
-
   const handleDateTimePickerError = error => {
     setDateTimePickerError(error);
   };
--- a/src/views/Rules/ListRules/index.jsx
+++ b/src/views/Rules/ListRules/index.jsx
@@ -27,6 +27,7 @@ import {
   getRules,
   getScheduledChanges,
   getScheduledChangeByRuleId,
+  addScheduledChange,
   deleteRule,
 } from '../../../services/rules';
 import { getRequiredSignoffs } from '../../../services/requiredSignoffs';
@@ -112,6 +113,7 @@ function ListRules(props) {
   );
   const fetchRequiredSignoffs = useAction(getRequiredSignoffs)[1];
   const delRule = useAction(deleteRule)[1];
+  const scheduleDelRule = useAction(addScheduledChange)[1];
   const [signoffAction, signoff] = useAction(props =>
     makeSignoff({ type: 'rules', ...props })
   );
@@ -356,9 +358,16 @@ function ListRules(props) {
 
   const handleDeleteDialogSubmit = async state => {
     const dialogRule = state.item;
-    const { error } = await delRule({
+    const now = new Date();
+    const when = scheduleDeleteDate >= now ? scheduleDeleteDate.getTime() : now.getTime() + 5000;
+    const { error } = Object.keys(dialogRule.required_signoffs).length === 0 ? await delRule({
       ruleId: dialogRule.rule_id,
       dataVersion: dialogRule.data_version,
+    }) : await scheduleDelRule({
+      change_type: 'delete',
+      when,
+      rule_id: dialogRule.rule_id,
+      data_version: dialogRule.data_version,
     });
 
     if (error) {
--- a/src/views/Rules/ListRules/index.jsx
+++ b/src/views/Rules/ListRules/index.jsx
@@ -1,4 +1,4 @@
-import React, { Fragment, useEffect, useState, useMemo } from 'react';
+import React, { Fragment, useEffect, useState, useMemo, useRef } from 'react';
 import { stringify, parse } from 'qs';
 import { addSeconds } from 'date-fns';
 import Spinner from '@mozilla-frontend-infra/components/Spinner';
@@ -105,6 +105,7 @@ function ListRules(props) {
   const [scrollToRow, setScrollToRow] = useState(null);
   const [roles, setRoles] = useState([]);
   const [signoffRole, setSignoffRole] = useState('');
+  const ruleListRef = useRef(null);
   const [products, fetchProducts] = useAction(getProducts);
   const [channels, fetchChannels] = useAction(getChannels);
   const [rules, fetchRules] = useAction(getRules);
@@ -302,6 +303,7 @@ function ListRules(props) {
           }),
     [productChannelFilter, rulesWithScheduledChanges]
   );
+
   const handleDateTimePickerError = error => {
     setDateTimePickerError(error);
   };
@@ -348,6 +350,7 @@ function ListRules(props) {
           return newRule;
         })
       );
+      ruleListRef.current.recomputeRowHeights();
       handleSnackbarOpen({
         message: `Rule ${result.rule_id} successfully scheduled`,
       });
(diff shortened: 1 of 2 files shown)


helfi92:
Changing the `key` of any React component will force the the component to be re-rendered. Would this fix the issue?

mozbhearsum:
> Changing the `key` of any React component will force the the component to be re-rendered. Would this fix the issue?

It doesn't look like that works here. I tried changing the RuleCard's key:
```
         <RuleCard
           className={classes.ruleCard}
-          key={rule.rule_id}
+          key={`${rule.rule_id}-${rule.scheduledChange ? rule.scheduledChange.sc_id : ''}`}
           rule={rule}
           onRuleDelete={handleRuleDelete}
           onSignoff={() => handleSignoff(rule)}
           onRevoke={() => handleRevoke(rule)}
         />
```

...but it doesn't cause the row height to be recalculated. I'm guessing this is because the `List` is what calls `getRowHeight`, and it has its own key that is ultimately set by https://github.com/bvaughn/react-virtualized/blob/master/source/Grid/defaultCellRangeRenderer.js#L58.

helfi92:
We use `VariableSizeList` to render the react-virtualized component. Did you try changing the key of`VariableSizeList`? Perhaps setting it to have `key={filteredRulesCount}`. My understanding is that it will force whatever is inside `VariableSizeList` to be re-created.

https://github.com/mozilla-frontend-infra/balrog-ui/blob/3de4510dccccaadb9a41519a715294840fbb30dd/src/views/Rules/ListRules/index.jsx#L696-L701

mozbhearsum:
> We use VariableSizeList to render the react-virtualized component. Did you try changing the key ofVariableSizeList? Perhaps setting it to have key={filteredRulesCount}. My understanding is that it will force whatever is inside VariableSizeList to be re-created.

It didn't occur to me that I could just add a key to an object outside of a loop :upside_down_face:. What do you think about setting the key to the md5 hash of `filteredRulesWithScheduledChanges`? The alternative would be concat'ing all of the rule ids and sc ids together, I think.

helfi92:
> What do you think about setting the key to the md5 hash of filteredRulesWithScheduledChanges?

In what circumstance will this hash change? Will deleting an item from filteredRulesWithScheduledChanges without re-fetching the list of rules make that value change?

mozbhearsum:
> > What do you think about setting the key to the md5 hash of filteredRulesWithScheduledChanges?
> 
> In what circumstance will this hash change? Will deleting an item from filteredRulesWithScheduledChanges without re-fetching the list of rules make that value change?

As far as I know, this will change whenever `productChannelFilter` or `rulesWithScheduledChanges` changes (https://github.com/mozilla-frontend-infra/balrog-ui/blob/1f952b3fe80334babd3e6e5275cc37ad8d9329a0/src/views/Rules/ListRules/index.jsx#L303)

Besides first load, `rulesWithScheduledChanges` changes when:
* A delete is done or scheduled (https://github.com/mozilla-frontend-infra/balrog-ui/blob/1f952b3fe80334babd3e6e5275cc37ad8d9329a0/src/views/Rules/ListRules/index.jsx#L329)
* A signoff is made or revoked (https://github.com/mozilla-frontend-infra/balrog-ui/blob/1f952b3fe80334babd3e6e5275cc37ad8d9329a0/src/views/Rules/ListRules/index.jsx#L400, https://github.com/mozilla-frontend-infra/balrog-ui/blob/1f952b3fe80334babd3e6e5275cc37ad8d9329a0/src/views/Rules/ListRules/index.jsx#L470)

The height will always change when we schedule a deletion (because the scheduled changes section will now be shown), and has the potential to for signoff/revoke. The height will never change for a direct deletion, however (the card is simply removed, instead).

helfi92:
Thanks for elaborating. Makes sense to me. Let's try going ahead with the md5 hash :)

mozbhearsum:
> Thanks for elaborating. Makes sense to me. Let's try going ahead with the md5 hash :)

It's looking like this may not be a great option. The full rerender is a little sluggish (especially noticable when signing off or revoking a signoff), and it also causes a scroll all the way back to the top.

helfi92:
> The full rerender is a little sluggish (especially noticable when signing off or revoking a signoff), and it also causes a scroll all the way back to the top.

Hm, in that case I think we could try the `ref` option. We might be able to use https://reactjs.org/docs/hooks-reference.html#useref to have a reference to the component.

mozbhearsum:
This was pretty straightforward since we're already setting a ref on `VariableSizeList`.

src/views/Rules/ListRules/index.jsx
```diff
@@ -707,6 +716,7 @@ function ListRules(props) {
                 scrollToRow={scrollToRow}
                 rowHeight={getRowHeight}
                 rowCount={filteredRulesCount}
+                listRef={ruleListRef}
```
helfi92: Can this be renamed to `ref` instead of `listRef`? I _think_ `ref` is like `key` where the child component can't access that value and wont overwrite props.

```jsx
<ButtonComponent key={9999} />

function ButtonComponent(props) {
    console.log(props.key) // prints undefined
}
```

mozbhearsum: Something like https://gist.github.com/mozbhearsum/3ed4f518e8c0d6e09ff3bc270ed31dca ?

I struggled a bit to make this work due to React things I don't quite understand. See https://github.com/facebook/flow/issues/7467, https://github.com/reactjs/rfcs/issues/103

helfi92: > Something like https://gist.github.com/mozbhearsum/3ed4f518e8c0d6e09ff3bc270ed31dca

Yes! Exactly that. Maybe I can help a bit:

`useRef`:
* is a hook
* common use case is to access a child methods
* can also be used as an instance variable in a class but that has nothing to do with what we're trying to do

In our case, we use `useRef` to have access to methods from the List component of react-virtualized.

`useImperativeHandle`:
* is a hook
* customizes the instance value (`ref` value) that is exposed to the parent when the parent decides to use `useRef` to have access to the child methods
* `useImperativeHandle` should be used with `forwardRef`

In our case, we let `VariableSizeList` take care of interacting with the internals of react-virtualized and expose a method via `useImperativeHandle` to ListRules that it can use.

`forwardRef`:
* Allows you to forward a `ref` prop down the React tree. React won’t pass some props down (e.g., `key` and `ref`) since they are considered somewhat special. `forwardRef` will have React allow it:

```jsx
// React will pass `ref` as a second parameter
function Child(props, ref) {

}
```

In our case, we use `forwardRef` to attach a method to that ref via `useImperativeHandle`.

mozbhearsum: Ah, so `forwardRef` is really just a flag to say "allow the ref to be passed"?

Thank you for the detailed explanation, this is very helpful.

mozbhearsum:
I realize this drops `propTypes` and `defaultProps`, but I think that's unavoidable due to https://github.com/facebook/flow/issues/7467 (unless we switch to a class based component).

src/views/Rules/ListRules/index.jsx
helfi92: Correct. No problem :)

helfi92:
> I realize this drops propTypes and defaultProps, but I think that's unavoidable due to facebook/flow#7467 (unless we switch to a class based component).

What happens when you try to add props? The linked issue has to do with prop-types not working with flow (fb static typing).

mozbhearsum:
> > I realize this drops propTypes and defaultProps, but I think that's unavoidable due to [facebook/flow#7467](https://github.com/facebook/flow/issues/7467) (unless we switch to a class based component).
> 
> What happens when you try to add props? The linked issue has to do with prop-types not working with flow (fb static typing).

Things still work, but we get this error in the console:
`Warning: forwardRef render functions do not support propTypes or defaultProps. Did you accidentally pass a React component?`

helfi92:
Hmm, I wonder if something like this would work:

```jsx
const WrappedComponent = React.forwardRef(
  function Component(props, ref) {
    return < ... />;
  }
);

WrappedComponent.propTypes = {
  // ...
};

WrappedComponent.defaultProps = {
  // ...
};
```

mozbhearsum:
Looks like it!

helfi92:
w00t! I will review soon :) Btw, WrappedComponent could just be `VariableSizeList` heh. Sorry for the confusion.

mozbhearsum:
> w00t! I will review soon :) Btw, WrappedComponent could just be `VariableSizeList` heh. Sorry for the confusion.

Oh haha, of course.

mozbhearsum:
Ah, I know what's up there. I'll fix it in a follow-up, thanks for the report!

Issue #111, opened 2019-07-23
Do something with the index page


mozbhearsum:
Right now it just says "Home". A very simple thing we could do is provide links to common filters of the Rules page. Eg: Firefox release channel and some others.

Something even nicer would be #88, but that may be a lot more work.

helfi92:
> A very simple thing we could do is provide links to common filters of the Rules page. Eg: Firefox release channel and some others.

That's a really good idea! We do something similar in Taskcluster UI:

### Logged out

![Screen Shot 2019-07-24 at 9 59 16 AM](https://user-images.githubusercontent.com/3766511/61799820-cbd97980-adf9-11e9-8a52-36b7893d72cb.png)


### Logged in

![Screen Shot 2019-07-24 at 9 58 06 AM](https://user-images.githubusercontent.com/3766511/61799740-a6e50680-adf9-11e9-93ad-125885f0b726.png)

helfi92:
![Screen Shot 2019-08-12 at 5 04 31 PM](https://user-images.githubusercontent.com/3766511/62898369-4cb5d200-bd23-11e9-9c6d-b75e57883516.png)

PR #146, opened 2019-08-12
Add a simple homepage with rules common filters

helfi92:
I wasn't sure which common filters to put. Open to suggestions.

Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/111.

### DIFF
--- a/src/utils/constants.js
+++ b/src/utils/constants.js
@@ -119,3 +119,25 @@ export const ALL_PERMISSIONS = [
   'scheduled_change',
 ];
 export const NEW_LINES_REGEX = /\r?\n|\r/g;
+export const RULES_COMMON_FILTERS = [
+  {
+    label: 'Firefox Release',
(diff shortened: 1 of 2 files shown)


helfi92:
Screen Shot 2019-08-12 at 5 04 31 PM

helfi92:
> Firefox DevEdition (http://localhost:9000/rules?product=Firefox&channel=aurora)

The text says DevEdition but the link goes to Aurora. Which one is correct.?

mozbhearsum:
> > Firefox DevEdition (http://localhost:9000/rules?product=Firefox&channel=aurora)
> 
> The text says DevEdition but the link goes to Aurora. Which one is correct?

Both :). DevEdition is the name of the product, but it uses the "aurora" channel for complicated historical reasons. It's not the end of the world if we use Aurora instead of DevEdition (everyone using Balrog understands they're the same).

Issue #147, opened 2019-08-12
Improve legibility of priority value on rules pages


nthomas-mozilla:
I'm finding it quite hard to read the priority for rules with the white text on light grey circle, it has much lower contrast than other parts of the UI, eg the 9000 in:
![Screen Shot 2019-08-13 at 9 04 05 AM](https://user-images.githubusercontent.com/942272/62898353-600b8080-bda9-11e9-81da-7fa532b0d4f2.png)

I am not good at UI colours but some suggestions - black on the existing gray is a bit better, but white-on-black or black-on-white could be worth a try. Or even something-on-banner-purple 😄

helfi92:
Agreed. It's a bit hard to read.

PR #148, opened 2019-08-12
Improve legibility of priority number in a rule card

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/147.

#### Before

Screen Shot 2019-08-12 at 5 33 09 PM

#### After

Screen Shot 2019-08-12 at 5 32 51 PM

### DIFF
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -108,6 +108,7 @@ const useStyles = makeStyles(theme => ({
   avatar: {
     height: theme.spacing(4),
     width: theme.spacing(4),
+    background: theme.palette.grey[500],
   },
   avatarText: {
     fontSize: 10,


mozbhearsum:
I'm going to defer to @nthomas-mozilla on this one.

nthomas-mozilla:
Definitely better but we could do more I think. How about:
Screen Shot 2019-08-15 at 12 19 18 PM

Brings in the primary color, increases priority text size, and uses an ellipsoid to minimize vertical size of the avatar. Priority is important to me, but might make it too prominent for the general case ? Very open to discussion, it's your project.

Generated with
```
diff --git a/src/components/RuleCard/index.jsx b/src/components/RuleCard/index.jsx
index 42459d1..fc31967 100644
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -107,11 +107,11 @@ const useStyles = makeStyles(theme => ({
   },
   avatar: {
     height: theme.spacing(4),
-    width: theme.spacing(4),
-    background: theme.palette.grey[500],
+    width: theme.spacing(6),
+    background: theme.palette.primary.main,
   },
   avatarText: {
-    fontSize: 10,
+    fontSize: theme.typography.body2.fontSize,
   },
   comment: {
     maxHeight: theme.spacing(10),
```

nthomas-mozilla:
Looks good to me, thanks Hassan.

PR #154, opened 2019-08-13
Allow full deletes of Required Signoffs, and cancelling deletions

mozbhearsum:
This implements one-click deletions for Required Signoffs via the SpeedDial button (you can already do this, but you have to click the trash icon on each signoff on the page).

While testing this, I also noticed we never wired up the "Cancel Delete" button on `ListSignoffs`, so I did that as well.

### DIFF
--- a/src/components/DialogAction/index.jsx
+++ b/src/components/DialogAction/index.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { oneOfType, object, node, string, func, bool } from 'prop-types';
 import { makeStyles } from '@material-ui/styles';
 import Dialog from '@material-ui/core/Dialog';
@@ -44,6 +44,12 @@ function DialogAction(props) {
     destructive,
     ...rest
   } = props;
+  let cancelled = false;
+
+  // Set this so we can avoid calling setActionExecuting if the caller
+  // decides to do something that unmounts us in onComplete.
+  useEffect(() => () => (cancelled = true));
+
   const handleSubmit = async () => {
     setActionExecuting(true);
 
@@ -63,7 +69,9 @@ function DialogAction(props) {
       onComplete(result);
     }
 
-    setActionExecuting(false);
+    if (!cancelled) {
+      setActionExecuting(false);
+    }
   };
 
   return (
--- a/src/components/DialogAction/index.jsx
+++ b/src/components/DialogAction/index.jsx
@@ -47,7 +47,7 @@ function DialogAction(props) {
   let cancelled = false;
 
   // Set this so we can avoid calling setActionExecuting if the caller
-  // decides to do something that unmounts us in onComplete.
+  // decides to do something that unmounts us in onComplete or onError.
   useEffect(() => () => (cancelled = true));
 
   const handleSubmit = async () => {
@@ -60,7 +60,9 @@ function DialogAction(props) {
         onError(error);
       }
 
-      setActionExecuting(false);
+      if (!cancelled) {
+        setActionExecuting(false);
+      }
 
       return;
     }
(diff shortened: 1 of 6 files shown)


src/views/RequiredSignoffs/ViewSignoff/index.jsx
```diff
@@ -165,8 +165,23 @@ function ViewSignoff({ isNewSignoff, ...props }) {
...
+      roles: [],
+      originalRoles,
+      additionalRoles: [],
+      isNewSignoff,
+    });
+
+    if (!error) {
+      props.history.push('/required-signoffs');
+    }
+  };
```
helfi92: Any chance we could have the delete preceded with a confirmation dialog?

mozbhearsum: Yeah, that's probably a good idea given that it's a pretty significant event. Will do.

mozbhearsum:
The latest changes work, but it's caused this error: `Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function`. I traced it down to the fact that https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/components/DialogAction/index.jsx#L66 happens after the redirect is initiated, but I haven't been able to come up with a way to fix it. If there's some way to do the redirect in `onClose` that would probably work, but I'm not sure it's possible to distinguish between a user initiated close and a post-complete close?

src/views/RequiredSignoffs/ViewSignoff/index.jsx
```diff
@@ -165,8 +178,37 @@ function ViewSignoff({ isNewSignoff, ...props }) {
     }
   };
 
-  // TODO: Add delete logic
-  const handleSignoffDelete = () => {};
+  const handleSignoffDelete = () =>
+    setDialogState({
+      ...dialogState,
+      open: true,
+      title: 'Delete Required Signoffs?',
```
helfi92: Should `Signoffs` be `Signoff`?

mozbhearsum: Personally, I don't think so. We're deleting the entire collection of signoff requirements (that is to say, all of the roles that are currently required), so I think the plural is appropriate.

mozbhearsum:
I'm not 100% certain this is what you had in mind, but it appears to work, and seems to make sense based on my limited knowledge of the cleanup functions.

src/components/DialogAction/index.jsx
```diff
@@ -44,6 +44,12 @@ function DialogAction(props) {
     destructive,
     ...rest
   } = props;
+  let cancelled = false;
+
+  // Set this so we can avoid calling setActionExecuting if the caller
+  // decides to do something that unmounts us in onComplete.
+  useEffect(() => () => (cancelled = true));
```
helfi92: Could you check if it still works if this is rewritten as `useEffect(() => () => (cancelled = true), []);` to not have the empty useEffect triggered on each render.

src/components/DialogAction/index.jsx
```diff
@@ -63,7 +69,9 @@ function DialogAction(props) {
       onComplete(result);
     }
 
-    setActionExecuting(false);
+    if (!cancelled) {
+      setActionExecuting(false);
+    }
```
helfi92: This is exactly what I had in mind :) We should probably also wrap the other `setActionExecuting` with the if statement.

mozbhearsum: Good point. We don't have any onError handlers that would unmount at the moment, but we certainly could in the future...

src/components/DialogAction/index.jsx
mozbhearsum: It looks like this doesn't work. With this form of it, I'm guessing the cleanup handler only exists during the first render, and then disappears.

helfi92: Hm, weird :( https://stackoverflow.com/a/55020668/6150432 suggests that it should run. Does it print a console log statement if you put one in the cleanup function?

mozbhearsum: I'm going to go ahead and merge this as is, but if there's something we should look at as a follow-up to improve this, I'm happy to do so.

Issue #152, opened 2019-08-13
implement read only/read write mutations


mozbhearsum:
Still stubbed out in https://github.com/mozilla-frontend-infra/balrog-ui/blob/582f74d86ce24272457035804f57d23c1430cbf6/src/views/Releases/ListReleases/index.jsx#L105

mozbhearsum:
Fixed in #167.

PR #157, opened 2019-08-14
Implement "Restore" button on ListRuleRevisions

mozbhearsum:
I also fixed a small bug where we redirect incorrectly after deleting a scheduled change on the `Rule` view.


src/views/Rules/ListRuleRevisions/index.jsx
```diff
@@ -99,8 +101,27 @@ function ListRuleRevisions(props) {
...
+      change_type: 'update',
+      when: new Date().getTime() + 5000,
+      data_version: revisions[0].data_version,
+      ...rowData,
+    });
+
+    if (!error) {
+      props.history.push(`/rules#scId=${data.data.sc_id}`);
+    }
+  };
```
helfi92: Should we show a confirmation dialog when restoring? Are there any bad consequences if a user clicks on restore by mistake?

mozbhearsum: All of the important channels are protected by signoffs, so nothing bad can happen directly. But nonetheless, I think this is a good idea. I can see this button getting clicked by accident when trying to click `View`.

src/views/Rules/ListRuleRevisions/index.jsx
```diff
@@ -82,8 +89,44 @@ function ListRuleRevisions(props) {
...
-  const handleRestoreClick = () => {};
+  const handleRestoreClick = revision => () =>
+    setDialogState({
+      ...dialogState,
+      open: true,
+      body: `Rule ${
+        revision.alias ? revision.alias : revision.rule_id
+      } will be reverted to data version ${revision.data_version}.`,
+      title: 'Revert rule?',
+      confirmText: 'Revert',
```
helfi92: Screen Shot 2019-08-16 at 10 17 00 AM

I think we should either change `Restore` to `Revert` in the table or have the dialog say `Restore Rule?` and have `Restore` as `confirmText` to make sure we are consistent.

src/views/Rules/ListRuleRevisions/index.jsx
```diff
@@ -155,6 +200,17 @@ function ListRuleRevisions(props) {
...
+        open={dialogState.open}
+        title={dialogState.title}
+        body={dialogState.body}
+        confirmText={dialogState.confirmText}
+        onSubmit={handleDialogSubmit}
+        onError={handleDialogError}
+        error={dialogState.error}
+        onComplete={handleDialogActionComplete}
+        onClose={handleDialogClose}
+      />
```
helfi92: When an error occurs, I am seeing an infinite spinner on the button.

Screen Shot 2019-08-16 at 10 19 10 AM

I think it was working on the previous patch so it was probably something introduced after that.

mozbhearsum: I'll look into this.

mozbhearsum: Okay, so it looks like this is actually caused by the changes to `DialogAction` from #154. I did some more digging, and it appears that as things stand now, we set `cancelled = true` in every render - so we never `setActionExecuting(false)`.

I tried setting `[]` as the `useEffect` dependency again, which fixes this problem, but causes the initial error again. I think this is because setting `cancelled = true` happens in one render, and `setActionExecuting` happens in another (so `cancelled` ends up `false` again).

I also using a `setState` for `cancelled` instead of a local variable, but that doesn't appear to have helped. It seems like we need a value that persists across renders, is set to `true` when the component unmounts, and for `handleSubmit` to notice that change between calling `onComplete` and checking the value of `cancelled`. I'm a bit out of my depth here, so any tips would be appreciate.

mozbhearsum: It looks like this is also causing problems when implementing the read only mutations for #152 - `setActionLoading` is never set to false, so subsequent opens of the dialog start the spinner already moving.

helfi92: I'll see how I can help. Worst case we revert the cleanup patch and fix that warning error later.

helfi92: Could you check if the following diff fixes it:

```diff
diff --git a/src/components/DialogAction/index.jsx b/src/components/DialogAction/index.jsx
index a6237a8..ab49127 100644
--- a/src/components/DialogAction/index.jsx
+++ b/src/components/DialogAction/index.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { oneOfType, object, node, string, func, bool } from 'prop-types';
 import { makeStyles } from '@material-ui/styles';
 import Dialog from '@material-ui/core/Dialog';
@@ -30,6 +30,7 @@ const useStyles = makeStyles(theme => ({
 function DialogAction(props) {
   const classes = useStyles();
   const [actionExecuting, setActionExecuting] = useState(false);
+  const cancelled = useRef(false);
   const {
     open,
     title,
@@ -44,14 +45,17 @@ function DialogAction(props) {
     destructive,
     ...rest
   } = props;
-  let cancelled = false;
 
   // Set this so we can avoid calling setActionExecuting if the caller
   // decides to do something that unmounts us in onComplete or onError.
-  useEffect(() => () => (cancelled = true));
+  useEffect(() => () => {
+    cancelled.current = true;
+  });
 
   const handleSubmit = async () => {
-    setActionExecuting(true);
+    if (cancelled.current === false) {
+      setActionExecuting(true);
+    }
 
     const [error, result] = await tryCatch(onSubmit());
 
@@ -60,7 +64,7 @@ function DialogAction(props) {
         onError(error);
       }
 
-      if (!cancelled) {
+      if (cancelled.current === false) {
         setActionExecuting(false);
       }
 
@@ -71,7 +75,7 @@ function DialogAction(props) {
       onComplete(result);
     }
 
-    if (!cancelled) {
+    if (cancelled.current === false) {
       setActionExecuting(false);
     }
   };
```

^ uses the ref as an instance variable. (third point of https://github.com/mozilla-frontend-infra/balrog-ui/pull/142#discussion_r313188177)

mozbhearsum: Ah, clearly I haven't interalized refs yet. Your patch works as long as I add an empty array as the second arg to `useEffect`.

mozbhearsum:
I think this is ready to go now.

src/views/Rules/ListRuleRevisions/index.jsx
helfi92: I tried removing the empty array and reverting or using the dialog elsewhere still seems to work. Any chance you could provide me with some STR of something that shows the dialog not working if the empty array is not specified?

mozbhearsum: My STR is:
1) Load http://localhost:9000/rules/145/revisions
2) Click "Restore" on a revision
3) Click "Revert" on the dialog (oops, I need to change that text still).
4) Observe that while the button does change colour (light purple to dark purple), it does not get disabled.

With the empty array, both the Cancel and the Revert button get greyed out.

src/views/Rules/ListRuleRevisions/index.jsx
```diff
@@ -71,8 +78,44 @@ function ListRuleRevisions(props) {
...
-  // TODO: Add logic to restore a revision
-  const handleRestoreClick = () => {};
+  const handleRestoreClick = revision => () =>
+    setDialogState({
+      ...dialogState,
+      open: true,
+      body: `Rule ${
+        revision.alias ? revision.alias : revision.rule_id
+      } will be restored to data version ${revision.data_version}.`,
+      title: 'Restore rule?',
```
helfi92: `s/rule/Rule/`

PR #160, opened 2019-08-15
Virtualize list of rule revisions

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/104.


helfi92:
I increased the `overscanRowCount` and added a `estimatedRowSize` prop. I think it's a little bit faster now.

PR #163, opened 2019-08-15
Fix parsing of release history username.

mozbhearsum:
I noticed the current parsing doesn't work correctly when the username contains a hyphen, which is a bug in the code we carried over from the old UI.

### DIFF
--- a/src/services/releases.js
+++ b/src/services/releases.js
@@ -24,7 +24,7 @@ const getRevisions = (name, product) => {
           product,
           data_version: Number(dataVersion) ? dataVersion : parts[0],
           timestamp: parseInt(parts[1], 10),
-          changed_by: parts[2],
+          changed_by: parts.slice(2).join('-'),
           data_url: r.mediaLink,
         };
 


src/services/releases.js
```diff
@@ -24,7 +24,7 @@ const getRevisions = (name, product) => {
           product,
           data_version: Number(dataVersion) ? dataVersion : parts[0],
           timestamp: parseInt(parts[1], 10),
-          changed_by: parts[2],
+          changed_by: parts.slice(2).join('-'),
```
helfi92: Out of curiosity, why do `parts.slice(2)` instead of `parts[2]`?

mozbhearsum: For usernames like `balrog-ffxbld`, we end up with more than 3 items in the Array, so we want to grab the 3rd item and on:
```
var a = "Devedition-68.0b1-build1/483-1557763308625-balrog-ffxbld.json"
undefined
a.replace(`Devedition-68.0b1-build1/`, '').replace('.json', '').split('-')[2]
"balrog"
a.replace(`Devedition-68.0b1-build1/`, '').replace('.json', '').split('-').slice(2).join("-")
"balrog-ffxbld"

var a = "Devedition-68.0b1-build1/483-1557763308625-bhearsum@mozilla.com.json"
undefined
a.replace(`Devedition-68.0b1-build1/`, '').replace('.json', '').split('-')
Array(3) [ "483", "1557763308625", "bhearsum@mozilla.com" ]
a.replace(`Devedition-68.0b1-build1/`, '').replace('.json', '').split('-').slice(2).join("-")
"bhearsum@mozilla.com"
```

helfi92: Got it, thanks!

Issue #59, opened 2019-07-08
be able to add/update releases


mozbhearsum:
Both directly and through scheduled changes

helfi92:
We could use https://mozilla-frontend-infra.github.io/components/#codeeditor for the JSON editor.

PR #164, opened 2019-08-17
Add update/create release view

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/59.


helfi92:
Thank you for the review :) I'll look into this as soon as I get the chance.

helfi92:
> Performance is really slow (at least on my laptop) for large blobs, eg: any Devedition or Firefox one. It takes 1-2s for anything I type to show up, and about the same amount of time to scroll horizontally.

I had to create a lighter `CodeEditor` component which makes it work better with a 15k lines of code release.

src/views/Releases/Release/index.jsx
```diff
@@ -1,11 +1,163 @@
...
+  const fetchReleases = useAction(getReleases)[1];
+  const [products, fetchProducts] = useAction(getProducts);
+  const isLoading = release.loading || products.loading;
+  // TODO: Fill actionLoading when hooking up mutations
+  const actionLoading = false;
+  const error = release.error || products.error;
+
+  useEffect(() => {
+    if (releaseName) {
+      Promise.all([fetchRelease(releaseName), fetchReleases()]).then(
```
mozbhearsum: It looks like the only purpose of `fetchReleases` is to pull out the name of the release to set `productTextValue`. Given that we already have it (in edit mode it's in the URL, in create mode it's user provided in `productTextValue`), is this necessary?

src/views/Releases/Release/index.jsx
```diff
@@ -1,11 +1,163 @@
...
+            );
+
+            setProductTextValue(r.product);
+          }
+        }
+      );
+    }
+
+    fetchProducts();
+  }, [releaseName]);
```
mozbhearsum: Given that `releaseName` is static (as far as I can tell, at least), does this need to depend on it? I'm guessing this serves a purpose that I'm not aware of, but because it makes XHR requests that don't need to be done more than once, I'm curious.

src/views/Releases/Release/index.jsx
helfi92: I don't see the product name in the URL. For example, localhost:9000/releases/Devedition-56.0b3-build1 only has the release name. Am i missing something?

src/views/Releases/Release/index.jsx
helfi92: It would probably still work in this case but we are strongly recommended to mention all of the dependencies that the effect depends on. This would avoid any weird errors like seeing a previous value instead of the latest one. From https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect:

> every value referenced inside the effect function should also appear in the dependencies array. In the future, a sufficiently advanced compiler could create this array automatically.

src/views/Releases/Release/index.jsx
mozbhearsum: Sorry, I don't know why I thought product could be in the URL. My brain kept thinking this was dealing with `name`.

src/views/Releases/Release/index.jsx
mozbhearsum: OK, makes sense. And since the dependency is effectively constant, this shouldn't ever refire?

src/views/Releases/Release/index.jsx
```diff
@@ -1,11 +1,163 @@
...
+  // TODO: Fill actionLoading when hooking up mutations
+  const actionLoading = false;
+  const error = release.error || products.error;
+
+  useEffect(() => {
+    if (releaseName) {
+      Promise.all([fetchRelease(releaseName), fetchReleases()]).then(
+        ([fetchedRelease, fetchedReleases]) => {
+          if (fetchedRelease.data) {
+            setReleaseEditorValue(JSON.stringify(fetchedRelease.data, null, 2));
```
mozbhearsum: (I thought I posted this already, but it appears to have disappeared).

I think this needs to be:
```suggestion
            setReleaseEditorValue(JSON.stringify(fetchedRelease.data.data, null, 2));
```

Currently, I'm seeing part of the axios response in this field.

src/views/Releases/Release/index.jsx
helfi92: Correct.

src/views/Releases/Release/index.jsx
helfi92: Oops. Just pushed a fix.

PR #166, opened 2019-08-20
Security header updates

mozbhearsum:
A few updates based on the Mozilla Observatory report (https://observatory.mozilla.org/analyze/balrog-admin-static-stage.stage.mozaws.net). I'd like to get rid of 'unsafe-inline' for style-src, but it looks like a pretty tricky affair, so I'm going to defer it for now.


.neutrinorc.js
```diff
@@ -26,19 +26,24 @@ module.exports = {
               img-src: gravatar & githubusercontent & i1.wp.com for user avatars
                        (which can't be locked down to a specific path due to CSP format)
               style-src: https://fonts.googleapis.com for dynamically loaded fonts
+                         'unsafe-inline' because styled components need them
```
helfi92: `s/styled components/CSS-in-JS/`

PR #170, opened 2019-08-20
Allow viewing of read only releases.

mozbhearsum:


src/components/CodeEditor/index.jsx
```diff
@@ -16,6 +16,7 @@ function CodeEditor({ onChange, value, ...rest }) {
         theme: 'material',
         indentWithTabs: false,
         lineNumbers: true,
+        readOnly,
```
helfi92: We can do `readOnly: readOnly ? 'nocursor' : false,` to remove the cursor as well. We'll then have to add this property to `propTypes` and `defaultProps`.

PR #171, opened 2019-08-21
Implement Release view mutations

mozbhearsum:
There's a few other small fixes/changes here too:
* Ensure `useAction` always passes along all positional arguments.
* Disable `Delete` button on `ReleaseCard` instead of hiding it when a Release has Rules pointing at it.
* Replace the `Delete Release` button on the `Release` view with `Cancel Pending Change` (like we have for Rules).

### DIFF
--- a/src/hooks/useAction.js
+++ b/src/hooks/useAction.js
@@ -17,12 +17,12 @@ export default action => {
   const [data, setData] = useState(null);
   const [error, setError] = useState(null);
   const state = { loading, data, error };
-  const performAction = async body => {
+  const performAction = async (...body) => {
     try {
       setLoading(true);
       setData(null);
       setError(null);
-      const data = await action(body);
+      const data = await action(...body);
 
       setData(data);
 
--- a/src/views/Releases/Release/index.jsx
+++ b/src/views/Releases/Release/index.jsx
@@ -14,7 +14,7 @@ import SpeedDial from '../../../components/SpeedDial';
 import AutoCompleteText from '../../../components/AutoCompleteText';
 import CodeEditor from '../../../components/CodeEditor';
 import useAction from '../../../hooks/useAction';
-import { getReleases, getRelease, createRelease } from '../../../services/releases';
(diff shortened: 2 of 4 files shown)


src/hooks/useAction.js
```diff
@@ -17,12 +17,12 @@ export default action => {
   const [data, setData] = useState(null);
   const [error, setError] = useState(null);
   const state = { loading, data, error };
-  const performAction = async body => {
+  const performAction = async (...body) => {
```
helfi92: 🥇

src/views/Releases/Release/index.jsx
```diff
@@ -152,11 +246,11 @@ export default function Release(props) {
             {!isNewRelease && (
```
helfi92: Similar to the rule view, let's now show the speed dial if the buttons are no changes to cancel.

src/views/Releases/Release/index.jsx
```diff
@@ -90,10 +130,64 @@ export default function Release(props) {
...
+    }
+  };
+
+  const handleReleaseUpdate = async () => {
+    // Updates to Releases require signoff in some circumstances,
+    // but we don't have a need to schedule them for specific times.
+    // Because of this, we always schedule them for slightly in the future,
+    // which means that changes that don't require signoff will happen
+    // almost immediately, and changes that do require signoff will wait
+    // until those are completed.
```
helfi92: The comments have been really helpful :)

mozbhearsum: Thanks! I hope my future self thinks so too :).

helfi92:
I think https://github.com/mozilla-frontend-infra/balrog-ui/pull/171#discussion_r316395641 still needs to be done.

src/views/Releases/Release/index.jsx
mozbhearsum: Oh, whoops. I forgot to push before merging, heh.

PR #174, opened 2019-08-22
Show scheduled changes on ListReleases and allow signoff/revoke

mozbhearsum:
This is a lot of copy/paste from RuleCard/ListRules to show scheduled changes and allow signoffs to happen. It's a little bit more complicated for Releases, because we now have 3 different modes the dialog can be in. I also found a bug that meant `handleSignoffDialogSubmit` was always using the previous value of `signoffRole` - which meant that you couldn't sign off with anything except the first role in the dialog on the first attempt.

I also still think there's something a bit off about the height calculations. Eg: body1 and body2 values don't appear to match up to what I see. I spent some time trying to improve this but it ended up being a rathole. This looks good enough at the moment for me, though it would be nice to fix at some point.


mozbhearsum:
Just fixed up the merge conflicts.

src/utils/constants.js
```diff
@@ -52,6 +52,7 @@ export const DIALOG_ACTION_INITIAL_STATE = {
   item: null,
   open: false,
   destructive: false,
+  mode: '',
```
helfi92: GIven that `mode` won't be inputted to a text field that expects a string, we could probably just make this `null`.

src/views/Releases/ListReleases/index.jsx
```diff
@@ -276,9 +474,76 @@ function ListReleases(props) {
...
+        const nSignoffs = Object.keys(release.scheduledChange.signoffs).length;
+        // Required Roles and Signoffs are beside one another, so we only
+        // need to account for the one with the most items.
+        const signoffRows = Math.max(requiredRoles, nSignoffs);
+
+        // Padding above the summary
+        height += theme.spacing(2);
+
+        // The "Requires Signoff From" title and the margin beneath it
+        height += listSubheaderTextHeight + theme.spacing(0.5);
```
helfi92: I think we are using an unusual subheader in the sense we are hard-coding the line-height. Instead, we could something along the lines of:

```jsx
height += signoffSummarylistSubheaderTextHeight + theme.spacing(0.5);

// elementsHeight.js
signoffSummarylistSubheaderTextHeight: remToPx(theme.typography.subtitle1.fontSize) * 1.5
```

Issue #149, opened 2019-08-12
Show more cards on screen


nthomas-mozilla:
🚨  bikeshed alert, whoop whoop whoop 🚨 

The new UI is more open because there is more whitespace in and around cards for rules and releases. This means less cards are visible on the screen at any one time, which makes it harder to consider several rules together. (On the flip side it's narrower, so it's possible to have two windows side by side). Ben told me that whitespace within a card is a bit painful to adjust, but inter-card spacing would be fairly straightforward. Worth a go ?

helfi92:
We could definitely try to minimize the space between cards and perhaps even make the card denser.

_Example: A lot of white space between some cards that I'm sure we can cut in half_

![Screen Shot 2019-08-12 at 6 17 25 PM](https://user-images.githubusercontent.com/3766511/62902395-74aa3300-bd2d-11e9-9e84-66e84fc64519.png)

> On the flip side it's narrower, so it's possible to have two windows side by side

This would entail moving away from using a dynamic list to using a multi grid that supports dynamic heights for cards. We currently use https://bvaughn.github.io/react-virtualized/#/components/List to have the list virtualized. If we do decide to go that route, we would need to look into using https://bvaughn.github.io/react-virtualized/#/components/MultiGrid so that things like scrolling to a given card on page load work as intended.

nthomas-mozilla:
Oh, I meant two browser windows, as a workaround. The single column layout has the nice property of making the rule-priority order clear, that might be trickier with a grid.

helfi92:
Wow, sorry for the misunderstanding. Yes, we can definitely reduce the spacing between cards :)

PR #180, opened 2019-08-28
Do not round result when converting rem to px

helfi92:
I think that'll fix it :)

Also relates to https://github.com/mozilla-frontend-infra/balrog-ui/issues/149.


helfi92:
Yay!

PR #184, opened 2019-08-30
Emergency Shutoff UI

mozbhearsum:
This probably needs a little bit more work, but I think it's far enough along for some feedback. Under normal circumstances, the only visible change is the SpeedDial with the "Disable Updates" button in the bottom right. Once updates have been shut off, the EmergencyShutoffCard is shown:
image

When "Enable Updates" is used for a Product/Channel that doesn't require signoff, we make a request to the backend to re-enable them, and remove the EmergencyShutoffCard if that's successful.

If signoffs are required, the card changes to the following:
image

I got rid of almost all of the information we show on other Scheduled Change sections (what will change, when it will change, etc.) because it's either not relevant, or shown in the CardHeader instead.

If "Keep Updates Disabled" is clicked, we cancel the enabling of updates, which reverts the Card back to the first screenshot. The "Signoff" button works as usual.


src/views/Rules/ListRules/index.jsx
```diff
@@ -817,6 +1170,16 @@ function ListRules(props) {
         onExited={handleDialogExited}
       />
       <Snackbar onClose={handleSnackbarClose} {...snackbarState} />
+      <SpeedDial ariaLabel="Secondary Actions">
+        <SpeedDialAction
+          disabled={isLoading || !username || filteredProductChannelIsShutoff}
+          icon={<CloseNetworkIcon />}
```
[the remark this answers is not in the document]
mozbhearsum: I like this a lot!

src/views/Rules/ListRules/index.jsx
```diff
@@ -279,6 +364,39 @@ function ListRules(props) {
...
+        )
+      );
+    }
+  }, [requiredSignoffs, searchQueries]);
+
+  useEffect(
+    () =>
+      setFilteredProductChannelIsShutoff(
+        emergencyShutoffs.some(es => {
+          if (productChannelFilter === ALL) {
```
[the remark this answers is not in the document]
mozbhearsum: With apologies, I think I changed this underneath you in https://github.com/mozilla-frontend-infra/balrog-ui/pull/184/commits/52b249008703c1b60c1fee0a91433cb0a3883e9b - so I think it's no longer relevant.

src/views/Rules/ListRules/index.jsx
```diff
@@ -567,12 +686,226 @@ function ListRules(props) {
...
+    setEmergencyShutoffs(
+      emergencyShutoffs.map(es => {
+        if (
+          !es.scheduledChange ||
+          es.scheduledChange.sc_id !== emergencyShutoff.scheduledChange.sc_id
+        ) {
+          return es;
+        }
+
+        const newEs = { ...es };
```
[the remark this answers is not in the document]
mozbhearsum: Sounds good. What's the difference between the two, anyways? Deep vs. shallow clone?

src/views/Rules/ListRules/index.jsx
```diff
@@ -817,6 +1170,16 @@ function ListRules(props) {
...
       <Snackbar onClose={handleSnackbarClose} {...snackbarState} />
+      <SpeedDial ariaLabel="Secondary Actions">
+        <SpeedDialAction
+          disabled={isLoading || !username || filteredProductChannelIsShutoff}
+          icon={<CloseNetworkIcon />}
+          className={classes.disableUpdates}
+          tooltipOpen
+          tooltipTitle="Disable Updates"
+          onClick={handleDisableUpdates}
+        />
```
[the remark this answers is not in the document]
mozbhearsum: Ah, good catch. Looks like this needs to be done when only a product is filtered, too.

src/components/EmergencyShutoffCard/index.jsx
```diff
@@ -0,0 +1,204 @@
...
+            (user && user.email in emergencyShutoff.scheduledChange.signoffs ? (
+              <Button color="secondary" disabled={!user} onClick={onRevoke}>
+                Revoke Signoff
+              </Button>
+            ) : (
+              <Button color="secondary" disabled={!user} onClick={onSignoff}>
+                Signoff
+              </Button>
+            ))}
+        </CardActions>
```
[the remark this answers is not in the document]
mozbhearsum: Hm, it looks like we don't have an existing pattern to do this in components (we only do it directly in the `Rule` and `Release` view at the moment). Can we punt this and deal with it more generally elsewhere?

mozbhearsum:
I've addressed most of the comments - let me know how you feel about the open threads above.

src/views/Rules/ListRules/index.jsx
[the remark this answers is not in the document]
helfi92: > Deep vs. shallow clone?

Yep!

src/components/EmergencyShutoffCard/index.jsx
```diff
@@ -0,0 +1,128 @@
...
+      '&:hover .anchor-link-style': {
+        display: 'inline-block',
+        opacity: 1,
+        color: theme.palette.text.hint,
+        '&:hover': {
+          color: theme.palette.text.secondary,
+        },
+      },
+    },
+  },
```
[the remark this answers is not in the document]
mozbhearsum: This is nice - I like how it calls it out as different, and more important, than the other cards.

src/components/EmergencyShutoffCard/index.jsx
```diff
@@ -0,0 +1,113 @@
...
+    <Card classes={{ root: classes.root }} spacing={4} {...props}>
+      <CardHeader
+        classes={{ avatar: classes.cardHeaderAvatar }}
+        className={classes.cardHeader}
+        avatar={<AlertIcon />}
+        title={
+          <Typography component="h2" variant="h6">
+            {emergencyShutoff.scheduledChange
+              ? 'Updates are scheduled to be enabled for this product and channel'
+              : 'Updates are currently disabled for this product and channel'}
```
helfi92: Instead of `for this product and channel`, it would be great if we could substitute the actual names there.

src/views/Rules/ListRuleRevisions/index.jsx
```diff
@@ -48,9 +48,7 @@ function ListRuleRevisions(props) {
   const [fetchedRevisions, fetchRevisions] = useAction(getRevisions);
   const addSC = useAction(addScheduledChange)[1];
   const { ruleId } = props.match.params;
-  // eslint-disable-next-line prefer-destructuring
-  const error = fetchedRevisions.error;
-  const isLoading = fetchedRevisions.loading;
+  const { error, loading: isLoading } = fetchedRevisions;
```
helfi92: This was actually intended so that if we ever add new requests to this view, we could just do

```jsx
const error = fetchedRevisions.error || ...
const isLoading = fetchedRevisions.loading || ...;
```

src/views/Rules/ListRules/index.jsx
```diff
@@ -780,6 +1087,27 @@ function ListRules(props) {
...
+                className={classes.card}
+                emergencyShutoff={emergencyShutoffs.find(
+                  es =>
+                    es.product === searchQueries[0] &&
+                    es.channel === searchQueries[1]
+                )}
+                onEnableUpdates={handleEnableUpdates}
+                onCancelEnable={handleCancelEnableUpdates}
+                onSignoff={handleSignoffEnableUpdates}
+                onRevoke={handleRevokeEnableUpdates}
```
helfi92: Do you mind showing showing how the card looks when there are required signoffs, please? I tried adding more signoffs required in `/required-signoffs/<product>/<channel>` but that didn't work. I'm probably doing something wrong. I actually just want to check that the spacing between the yellow banner and the subtitle looks good.

src/views/Rules/ListRuleRevisions/index.jsx
mozbhearsum: Ah, let's maintain this, in that case.

src/views/Rules/ListRules/index.jsx
mozbhearsum: image

src/components/EmergencyShutoffCard/index.jsx
```diff
@@ -57,8 +58,8 @@ function EmergencyShutoffCard({
         title={
           <Typography component="h2" variant="h6">
             {emergencyShutoff.scheduledChange
-              ? 'Updates are scheduled to be enabled for this product and channel'
-              : 'Updates are currently disabled for this product and channel'}
+              ? `Updates are scheduled to be enabled for the ${product} ${channel} channel`
+              : `Updates are currently disabled for the ${product} ${channel} channel`}
```
helfi92: I wonder what would be outputted when a filter has no channel (e.g., CDM)

src/views/Rules/ListRules/index.jsx
helfi92: 👌

src/components/EmergencyShutoffCard/index.jsx
mozbhearsum: Shouldn't be an issue, because the Emergency Shutoffs require a Product and Channel (arguably we should change this, but that's how it is now), and we don't enable the UI unless both are present for the current filter: https://github.com/mozilla-frontend-infra/balrog-ui/pull/184/files#diff-6bdfeacc9dead52092c38bbdc3958037R1153

Issue #190, opened 2019-09-04
Show login button if login has expired


helfi92:
Right now, you have to manually log out and log in, which seems like it should be unnecessary...

PR #191, opened 2019-09-04
Log user out if credentials have expired

helfi92:
This will make sure `userSession` is always the up-to-date value instead of relying on the value of the render (hook).

Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/190.


helfi92:
> Even after opening the localhost instance, closing it for many minutes, and reopening it - I still end up with a valid token.

I actually get logged out on my machine if I close the tab for more than 30 seconds (current expiration time). Is it possible you had another Balrog UI tab open that was refreshing the token by any chance?

mozbhearsum:
> > Even after opening the localhost instance, closing it for many minutes, and reopening it - I still end up with a valid token.
> 
> I actually get logged out on my machine if I close the tab for more than 30 seconds (current expiration time). Is it possible there was another Balrog UI tab open that was refreshing the token by any chance?

I just tried again (and made sure I only had one tab open) and the same thing occurred. I wonder if this is Silent Authentication working? If I understand `checkSession` correctly from https://github.com/auth0/auth0.js#auth0webauth, it should work as long as the auth0 session is still valid, regardless of the status of the current token.

helfi92:
> If I understand checkSession correctly from https://github.com/auth0/auth0.js#auth0webauth, it should work as long as the auth0 session is still valid, regardless of the status of the current token

You're absolutely right here. I think we can simply rely on the `onError` callback that gets triggerd if `checkSession` fails to log the user out. We don't need to look if the credentials have expired. `react-auth0-components` does that for us.

Issue #192, opened 2019-09-05
Updating a rule OS version text field


helfi92:
<img width="463" alt="Screen Shot 2019-09-05 at 12 01 05 PM" src="https://user-images.githubusercontent.com/3766511/64358693-ee53da80-cfd4-11e9-820b-c62f8e199db3.png">

It would be nice to have the OS version displayed in a multi-line text-field to improve readability.

mozbhearsum:
:+1: 

I think folks were also hoping that we could wrap the multiline OS Version in Rules such as:
![image](https://user-images.githubusercontent.com/49649/64378822-8b296e80-cffb-11e9-811c-3edacfc96e65.png)

PR #198, opened 2019-09-06
Improve user experience of OS version

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/192.

Screen Shot 2019-09-06 at 4 02 05 PM

Screen Shot 2019-09-06 at 4 02 32 PM

### DIFF
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -130,7 +130,7 @@ const useStyles = makeStyles(theme => ({
   link: {
     ...theme.mixins.link,
   },
-  scheduledPriorityChange: {
+  schedPriorityChange: {
     backgroundColor: '#c0dc91',
   },
 }));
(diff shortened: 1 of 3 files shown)


helfi92:
> And I don't know if it's a huge pain in the butt or not, but allowing multiple entries on one line (and only using a second one if we'd start scrolling) would greatly improve the usage of space. Particularly for locales, a lot of horizontal space is wasted because they're only a few characters each. I'm thinking the wrapping would work like this in an ideal world:

Hm, I feel it's a bit unconventional to allow two separators (comma and new lines) which might add code complexity and confusion to users. I guess we can keep the current way and probably revisit if we hear complaints.

mozbhearsum:
I'm not going to block on the comma and newline thing, but I'd like to see the first 3 points addressed:
* Put locale and osVersion beside one another
* Move them below the Version row again. I know this makes it slightly less visually appealing, but it maintains the sort-by-frequency-of-use ordering, which is more important IMO.
* Increase the number of rows.

I can drive this to completion, if you're tight on time.

helfi92:
I failed to push the commit 😅 One sec.

helfi92:
Done.

Issue #139, opened 2019-08-08
yarn install throws with node 12


helfi92:
From IRC:

> fyi, I had to pin fsevents to 1.2.9 so that yarn+node12 could setup the new ui

Possible fix: Lock the version of node using https://docs.npmjs.com/files/package.json#engines until there is a better solution than pining fsevents to 1.2.9.

mozbhearsum:
Let's be sure to update https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/docker-compose.yml#L7 to use node10 if we lock the node version.

PR #199, opened 2019-09-06
Lock node version

helfi92:
Resolves https://github.com/mozilla-frontend-infra/balrog-ui/issues/139.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,6 @@ services:
     build:
       context: .
       dockerfile: Dockerfile
-    image: node:10.16.3
     ports:
       - "9000:9000"
     environment:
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     build:
       context: .
       dockerfile: Dockerfile
-    image: node:current
+    image: node:10.16.3
```
mozbhearsum: Hm, I just noticed that we specify both `build` and `image` here, and `build` takes precedence. We should actually drop this, and update https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/Dockerfile#L1.

Can we pin to `node:10` instead, too? That way we'll pick up minor updates.

PR #202, opened 2019-09-09
Reduce spacing between cards

helfi92:
Fixes https://github.com/mozilla-frontend-infra/balrog-ui/issues/149.

### DIFF
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -116,10 +116,6 @@ const useStyles = makeStyles(theme => ({
   avatarText: {
     fontSize: theme.typography.body2.fontSize,
   },
-  comment: {
-    maxHeight: theme.spacing(10),
-    overflowY: 'auto',
-  },
   propertyWithScheduledChange: {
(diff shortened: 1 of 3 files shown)


src/components/RuleCard/index.jsx
```diff
@@ -645,11 +641,14 @@ function RuleCard({
...
                     <ListItemText
-                      className={classes.comment}
+                      title={rule.comment}
                       primaryTypographyProps={{
                         component: 'div',
                         className: classes.primaryText,
                       }}
+                      secondaryTypographyProps={{
+                        className: classes.textEllipsis,
+                      }}
```
helfi92: Most cards (at least in development) have single line comments. Optimizing for a single line would reduce the spacing between cards since we gave 3 lines of height to each card which made the spacing between cards without comments really big. This patch will show `...` when a comment is > 1 line. Hovering on the comment should show the full comment.

mozbhearsum: I think a tooltip is probably enough. We can look into a better solution later if it turns out not to be.

PR #207, opened 2019-09-12
Sort Rules based on scheduled change priority (if it exists)…

mozbhearsum:
…rather than plain Rule priority.

### DIFF
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -130,6 +130,9 @@ const useStyles = makeStyles(theme => ({
   link: {
     ...theme.mixins.link,
   },
+  scheduledPriorityChange: {
+    backgroundColor: '#c0dc91',
+  },
 }));
 
 function RuleCard({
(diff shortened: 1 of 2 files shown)


src/components/RuleCard/index.jsx
```diff
@@ -185,14 +194,14 @@ function RuleCard({
...
                 <Avatar
                   title="Priority"
                   aria-label="Priority"
                   className={classes.avatar}>
                   <Typography className={classes.avatarText}>
-                    {rule.priority}
+                    {headerPriority}
                   </Typography>
                 </Avatar>
                 {diffedProperties.includes('priority') &&
```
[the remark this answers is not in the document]
mozbhearsum: I think this is a good improvement, let's do it :)

src/components/RuleCard/index.jsx
```diff
@@ -185,25 +201,19 @@ function RuleCard({
...
               <Fragment>
                 <Avatar
-                  title="Priority"
-                  aria-label="Priority"
-                  className={classes.avatar}>
+                  title={priorityTitle}
+                  aria-label={priorityTitle}
+                  className={classNames(
+                    classes.avatar,
+                    classes.scheduledPriorityChange
```
helfi92: I think we only want to add `scheduledPriorityChange` if we're showing a scheduled change priority. Something like:

```jsx
classNames(
  classes.avatar,
  { [classes.scheduledPriorityChange]: isScheduledChange }
)
```

mozbhearsum: Whoops, sorry about that!

Issue #205, opened 2019-09-10
make it possible to view a Release directly from the Rules view


mozbhearsum:
It's common for us to want to see details about a Release when analyzing the Rules. In the old UI, we had "View" buttons that would show a modal window with the Mapping or Fallback Mapping's full data. I think we should do something similar here, perhaps by showing it in a Drawer?

helfi92:
Sounds good to me.

PR #211, opened 2019-09-18
Add "View Release" button

helfi92:
Resolves https://github.com/mozilla-frontend-infra/balrog-ui/issues/205.


helfi92:
> I'd be willing to forego syntax highlighting/colour if it makes a big difference.

It does a big difference! I sent a patch to remove the syntax highlighting.

> Let's add this for "Fallback Mapping", too.

Done.

Issue #216, opened 2019-09-20
consider a dedicated download button for Balrog releases page


lundjordan:
This was available on the old UI and is handy when needing to edit a release blob.

I tried hitting "update" then copy/pasting the json from within the text box but on my mac, <command>a to select and then <command>c to copy did not work in FF. The only other option is to select text with mouse and drag to bottom. This would take minutes to do with big release blobs and without tweaking scroll speed.

helfi92:
I can replicate on a release that is "Read-Only". Otherwise, "Command+A" works for me on a mac in FF.

lundjordan:
thanks for testing. maybe that was it. I was trying to copy the main ff release official rule which was probably read-only.

also, I neglected to release that github's markdown would scrub me writing "\< command \> a". Sounds like you figured it out.

helfi92:
Resolved in https://github.com/mozilla-frontend-infra/balrog-ui/commit/03121e5f464a80da43e3c7475db0e29c0ca6cafc.

lundjordan:
wfm. thanks!

PR #219, opened 2019-09-23
Add ability to COMMAND+A (select-all) read-only code

helfi92:
Relates to https://github.com/mozilla-frontend-infra/balrog-ui/issues/216.


mozbhearsum:
This looks good, I'm not sure we should close #216 immediately though. Let's keep it open and see how people feel about this as a solution? I have a feeling that a `Download` button still may be desired.

lundjordan:
confirmed, it wfm! thanks!

Issue #88, opened 2019-07-17
make it easy to see all pending signoffs for a role


mozbhearsum:
This is not necessary for an MVP, but a use case that has come up a few times is the ability to see what is what changes are awaiting signoff by a particular Role. This can be helpful during release days when we're trying to make sure everything that needs to be, is signed off on.

I can see this having a few possible forms:
* A user specific view, that shows the logged in user anything they could potentially sign off on.
* A role specific view, that shows anything that a given role could potentially sign off on.
* Something integrated into the existing Rule, Release, Required Signoffs, and User views, which somehow makes it very obvious which items are awaiting signoff.

Or it could be something entirely different.

mozbhearsum:
As a relatively simple thing to solve the immediate use case of 'what-in-the-heck-needs-signoff', how about a drop down like:
![image](https://user-images.githubusercontent.com/49649/65360205-f878fa00-dbcc-11e9-82a1-7c82a7d5e285.png)

I'm tentatively calling this "Display Mode", and would additionally filter the list based on the selected item.

cc'ing a few folks that have may have opinions @lizzard, @lundjordan, @rehandalal.

I still think something else may be useful, but I want to make sure we address the immediate need quickly.

lizzard:
That looks good to me!

helfi92:
Instead of a dropdown, how do you feel about simply adding a switch toggle component with a label along the lines of "Show rules requiring signoff"? If it's important for a user to sometimes view rules that have no signoffs required, then a dropdown like the image above sounds good to me.

helfi92:
> a name for the toggle that wouldn't cause confusion

Perhaps "Filter by rules requiring signoff" :)

lundjordan:
> Perhaps "Filter by rules requiring signoff" :)

A filter or toggle wfm. so long as it was also in the URL so I can share it quickly.

mozbhearsum:
I did a bit of hacking on this in https://github.com/mozilla-frontend-infra/balrog-ui/compare/master...mozbhearsum:display-mode-rules?expand=1. It looks like I'm going to be busy with other things for a week or two though, so I won't be able to finish it up immediately.

helfi92:
I may have some time this week to look into this. I'll be in PTO the week after. I'll see if I can continue on this, of course if that's alright with you.

PR #221, opened 2019-09-26
Make it easy to see all pending signoffs

helfi92:
Resolves https://github.com/mozilla-frontend-infra/balrog-ui/issues/88.

### DIFF
--- a/src/views/Rules/ListRules/index.jsx
+++ b/src/views/Rules/ListRules/index.jsx
@@ -131,7 +131,8 @@ function ListRules(props) {
   const [rulesWithScheduledChanges, setRulesWithScheduledChanges] = useState(
     []
   );
-  const [onlyShowPendingSignoffs, setOnlyShowPendingSignoffs] = useState(false);
+  const onlySignoffsQuery = Boolean(query.onlySignoffs);
+  const [onlyShowPendingSignoffs, setOnlyShowPendingSignoffs] = useState(onlySignoffsQuery);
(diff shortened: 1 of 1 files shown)


src/views/Rules/ListRules/index.jsx
```diff
@@ -398,45 +432,45 @@ function ListRules(props) {
...
-            }
+  const filteredRulesWithScheduledChanges = useMemo(() => {
+    let filteredRules = clone(rulesWithScheduledChanges);
 
-            return true;
-          }),
-    [searchQueries, productChannelFilter, rulesWithScheduledChanges]
-  );
+    // Pending signoff switch
+    if (filteredRules && Boolean(query.onlySignoffs)) {
```
mozbhearsum: I hadn't considered that we didn't need to have local state for this. Nice.

src/views/Rules/ListRules/index.jsx
```diff
@@ -398,45 +432,45 @@ function ListRules(props) {
...
+      filteredRules = filteredRules.filter(rule => rule.scheduledChange);
```
mozbhearsum: I think we need to do one of two things, either:
a) Change this to filter on rules where the length of `rule.scheduledChange.signoffs` is less than the sum of required signoffs
b) Relabel the toggle/query ary to indicate that we're filtering on scheduled changes, not just things requiring signoff.

Other than that, this patch looks great.

PR #224, opened 2019-09-28
Remove close button from ErrorPanel when onClose is null

allan-silva:
While I was working on frontend side of mozilla/balrog/pull/983, realized I will need this text information. Probably will be used in others pages later.

image


src/components/SignoffsRequiredText/index.jsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <Typography component="p" className={classes.requireSignoffWarning}>
+      This changes requires signoffs:&nbsp;
+      {Object.entries(requiredSignoffs)
+        .map(([role, count]) => `${count} from ${role}`)
+        .join(', ')}
+      .
+    </Typography>
+  );
+}
```
helfi92: I think we can just inline this in the view directly instead of creating a new component for it. The logic seems to be very minimal.

allan-silva:
@helfi92 When I apply the diff, the close button disappear from others error panels. Is this expected?

helfi92:
@allan-silva I ended up adding a `unclosable` prop to remove the close button. Let me know if you had any questions/concerns.

mozbhearsum:
I'll have a look at this today or tomorrow.

allan-silva:
@helfi92 sounds good for me, thanks!

allan-silva:
@mozbhearsum I like `MessagePanel`, `InfoPanel` is a `MessagePanel` with blue (maybe) background :smile:

mozbhearsum:
SGTM, let's do it :).

allan-silva:
@mozbhearsum @helfi92 , I changed the component name, but last days I have been very busy, so I do not test on all affected pages yet.
This is my first esperience with React, I'm not sure if this compilation message means that all is OK.

image

allan-silva:
Also, I realised that internally this component is designed to be a ErrorPanel. Should I change the internal impl too?

https://github.com/mozilla-frontend-infra/balrog-ui/blob/91fbe1999c1c2dfff0e481efd9209589533ff914/src/components/MessagePanel/index.jsx#L49-L60

mozbhearsum:
> @mozbhearsum @helfi92 , I changed the component name, but last days I have been very busy, so I do not test on all affected pages yet.
> This is my first experience with React, I'm not sure if this compilation message means that all is OK.
> 
> image

The `caniuse` message is ignorable, yeah. (We need to upgrade some packages in `package.json` at some point, though.)

I'll defer to @helfi92 on what to do about `ErrorPanel`.

mozbhearsum:
Ping - @helfi92?

helfi92:
The best way in my opinion would be to modify the [component upstream](https://mozilla-frontend-infra.github.io/components/#errorpanel). We would be able to rename ErrorPanel to something else, add a `variant` prop to select between "info", "error", and "warning". I suspect this may involve a lot of work, breaking changes, etc... The "lazy" way that I am leaning on is to simply change `balrog-ui/src/components/MessagePanel/index.jsx` which you seem to be doing. I don't think it's the end of the world if we have a component called MessagePanel but internally uses ErrorPanel :)

mozbhearsum:
Sounds good to me.

mozbhearsum:
@allan-silva - can you make @helfi92's suggested changes? Then we should be good to go here.

allan-silva:
@helfi92, I just renamed the prop "error" to "message" and don not added the "variant" prop, because "warning" on upstream component is a boolean value. Also for now MessagePanel can be error or warning only.

Makes sense to you?

Let me know if other changes are required.

src/components/MessagePanel/index.jsx
```diff
@@ -17,53 +17,65 @@ const useStyles = makeStyles(theme => ({
   },
 }));
 
-export default function ErrorPanel({
```
helfi92: Instead of re-writing & renaming the ErrorPanel, let's just create a separate MessagePanel. I created https://github.com/mozilla-frontend-infra/balrog-ui/pull/232 which we could use. @allan-silva Thoughts?

allan-silva: That is great, thanks!

allan-silva:
Closing in favor of #232

Issue #223, opened 2019-09-28
Consider not show Release data on Release update page as default


allan-silva:
The current Release data for `firefox-release` has 19067 lines. The Json file has around 800kB, in my tests it, freezes the Firefox when I try to upload a new Json file.

As sugestion, Release data should be shown only if user requests.

PR #225, opened 2019-09-29
Improve performance of editor

helfi92:
Resolves https://github.com/mozilla-frontend-infra/balrog-ui/issues/223.

Performance should be much much better now. Also disabled some buttons when user is logged out.

### DIFF
--- a/src/components/CodeEditor/index.jsx
+++ b/src/components/CodeEditor/index.jsx
@@ -3,6 +3,7 @@ import { string, func, bool } from 'prop-types';
 import { Controlled as CodeMirror } from 'react-codemirror2';
 import 'codemirror/lib/codemirror.css';
 import 'codemirror/theme/material.css';
+import 'codemirror/mode/javascript/javascript';
 
 function CodeEditor({ onChange, value, readOnly, ...rest }) {
   return (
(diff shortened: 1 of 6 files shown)


src/components/ReleaseCard/index.jsx
```diff
@@ -300,7 +300,7 @@ function ReleaseCard(props) {
       </CardContent>
       <CardActions className={classes.cardActions}>
         <Link className={classes.link} to={`/releases/${release.name}`}>
-          <Button color="secondary">
+          <Button color="secondary" disabled={!user}>
```
mozbhearsum: I think this button should stay enabled even if the user is not logged in. However, it should always read "View" in that case.

src/components/CodeEditor/index.jsx
```diff
@@ -3,6 +3,7 @@ import { string, func, bool } from 'prop-types';
 import { Controlled as CodeMirror } from 'react-codemirror2';
 import 'codemirror/lib/codemirror.css';
 import 'codemirror/theme/material.css';
+import 'codemirror/mode/javascript/javascript';
```
mozbhearsum: Is this what's doing the virtualization? I can't find much information on how codemirror works (eg: https://codemirror.net/mode/javascript/index.html is pretty sparse).

src/components/ReleaseCard/index.jsx
helfi92: Ah, makes sense. I think we can keep the current behavior of showing "Update" and have the actions in that view disabled. It will tell the user that updating a release is possible.

src/components/CodeEditor/index.jsx
helfi92: Virtualization works based on the height of the editor. Showing all of the editor was not virtualizing anything since the full DOM was displayed.

PR #226, opened 2019-10-03
Hide unused fields for Guardian on Rule Edit page.

mozbhearsum:
We're in the process of adding support for the new Guardian product to Balrog. One of the things that will be different about its updates, is that the update URL will not contain most of the fields that Firefox updates do. In order to avoid confusion, I'd like to hide these fields to avoid any confusion.

At this time, the backend is not going to be smart enough to refuse to accept Rules for Guardian that contain these fields, so I'm specifically _not_ hiding them on the Rules view. On the off chance they get set by accident, it would be confusing to have values but not show them.

### DIFF
--- a/src/views/Rules/Rule/index.jsx
+++ b/src/views/Rules/Rule/index.jsx
@@ -36,7 +36,7 @@ import { withUser } from '../../../utils/AuthContext';
 import {
   EMPTY_MENU_ITEM_CHAR,
   SPLIT_WITH_NEWLINES_AND_COMMA_REGEX,
-  RULE_PRODUCT_PROPERTIES,
+  RULE_PRODUCT_UNSUPPORTED_PROPERTIES,
 } from '../../../utils/constants';
 
 const initialRule = {
@@ -364,6 +364,12 @@ function Rule({ isNewRule, user, ...props }) {
     return `Update Rule ${ruleId}${rule.alias ? ` (${rule.alias})` : ''}`;
   };
 
+  const showField = field =>
+    !(
+      rule.product in RULE_PRODUCT_UNSUPPORTED_PROPERTIES &&
+      RULE_PRODUCT_UNSUPPORTED_PROPERTIES[rule.product].includes(field)
+    );
+
   return (
     <Dashboard title={getTitle()}>
       {isLoading && <Spinner loading />}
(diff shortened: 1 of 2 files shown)


helfi92:
I am back from PTO :) I will review this as soon as I can.

helfi92:
@mozbhearsum How can I test this? I don't seem to see "Guardian" in the list of rules.

mozbhearsum:
Ah, you'll need to create a rule for Guardian. It can be basically whatever you want (point at a Firefox Release or whatever - doesn't matter). Let me know if you have any trouble with that.

mozbhearsum:
> When I try to create a new rule with product "Guardian" , I am seeing an error panel with message update_type cannot be set to null/empty when scheduling insertion of a new rule. Any idea why this is happening?

Hm. I'm unable to reproduce this - it always gets set to `minor` for me. What exactly is your STR?

> Would it try to set an OS version even though it's a restricted property?

Great question. I will double check, and make sure we don't pass it to the backend.

helfi92:
> Hm. I'm unable to reproduce this - it always gets set to minor for me. What exactly is your STR?

Works now after running `docker-compose build` :)

helfi92:
> Great question. I will double check, and make sure we don't pass it to the backend.

Sounds good!

mozbhearsum:
I actually managed to repro the `update_type` issue as well - which was caused by `update_type` being considered an unsupported field. It is, technically (we ignore it on the backend for Guardian), but it's required by the database schema, so for now we can't totally ignore it :(. We can improve this later IMO.

mozbhearsum:
> > Would it try to set an OS version even though it's a restricted property?
> 
> Ah, it turns out we are still sending the other properties. I added a console statement in `handleCreateRule` and was able to see
> 
> Screen Shot 2019-10-10 at 4 55 30 PM
> 
> The osVersion is unsupported for the Guardian product but was set. We should probably fix that.

Are you sure? The call to `filterProductData` happens within the call to `updateSC`, so if you're just logging `data` you'd see stale contents. I tested these scenarios and wasn't able to get `osVersion` set - but I may have done something wrong!

mozbhearsum:
(see https://github.com/mozilla-frontend-infra/balrog-ui/pull/226/files#diff-b6d56a3781defa9b6101ab1eaa826383R242)

helfi92:
Oh, sorry about that. I had the log statement above filterProductData. I'll check again.

src/views/Rules/Rule/index.jsx
```diff
@@ -181,6 +182,13 @@ function Rule({ isNewRule, user, ...props }) {
     }
   };
 
+  const productSupportsField = field =>
+    !(
+      rule.product in RULE_PRODUCT_UNSUPPORTED_PROPERTIES &&
+      RULE_PRODUCT_UNSUPPORTED_PROPERTIES[rule.product].includes(field)
+    );
+  const filterProductData = data =>
+    pick(Object.keys(data).filter(field => productSupportsField(field)), data);
```
helfi92: `.filter(field => productSupportsField(field))` -> `.filter(productSupportsField)`

PR #228, opened 2019-10-09
Convert jaws/mig64 null values to EMPTY_MENU_ITEM_CHAR

### LINKED ISSUE #227: jaws & mig64 are set incorrectly when duplicating a Rule
I noticed today that when I used "Duplicate" to create a new Rule that jaws & mig64 ended up being set to false, even though the Edit UI showed them as "-". This doesn't happen with the plain "Add" button nor when editing a Rule.

mozbhearsum:
This feels like a horrible hack, but it should fix #227. I'm totally open to a better solution :).

### DIFF
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -784,7 +784,9 @@ function RuleCard({
             <Link
               className={classes.link}
               to={{
-                pathname: `/rules/${rule.rule_id}/duplicate`,
+                pathname: rule.rule_id
+                  ? `/rules/duplicate/ruleId/${rule.rule_id}`
+                  : `/rules/duplicate/scId/${rule.scheduledChange.sc_id}`,
                 state: {
                   rulesFilter,
                 },
--- a/src/components/RuleCard/index.jsx
+++ b/src/components/RuleCard/index.jsx
@@ -784,9 +784,8 @@ function RuleCard({
             <Link
               className={classes.link}
               to={{
-                pathname: '/rules/create',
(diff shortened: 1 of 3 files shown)


mozbhearsum:
The idea being that we'd pull the rule info from the backend instead of providing it as state? And I assume we'd use the same actual view code (`views/Rule/index.jsx`) and just add a new route?

helfi92:
Correct!

src/components/RuleCard/index.jsx
```diff
@@ -784,9 +784,8 @@ function RuleCard({
             <Link
               className={classes.link}
               to={{
-                pathname: '/rules/create',
+                pathname: `/rules/${rule.rule_id}/duplicate`,
                 state: {
-                  rule,
                   rulesFilter,
                 },
```
helfi92: Screen Shot 2019-10-10 at 2 11 15 PM

Hmm, clicking "Duplicate" for ^ takes the user to http://localhost:9000/rules/undefined/duplicate. Is there a workaround that comes to your mind to fix this issue?

mozbhearsum: Hm, I didn't realize we had the duplicate button for scheduled additions! The most obvious way to fix this to me is to have two new routes: /rules/duplicate/ruleId/:ruleId and /rules/duplicate/scId/:scId (or something like that). I'll see if I can come up with a better way, though.

mozbhearsum: This turns out to be a pretty good fix, I think, since we already handle ruleId/scId in the params.

PR #232, opened 2019-11-08
Add MessagePanel component

helfi92:


src/components/MessagePanel/index.jsx
```diff
@@ -0,0 +1,97 @@
...
+    info: InformationIcon,
+  };
+  const Icon = variantIcon[variant];
+  const handleClose = () => {
+    setDisplay(false);
+  };
+
+  if (!display) {
+    return null;
+  }
```
mozbhearsum: TIL you can do this.

PR #238, opened 2019-12-14
Replaced with default port.

divyanshu-rawat:
Project default pot defined as defined in` neutrinorc.js`
`const DEFAULT_PORT = 9000;`


divyanshu-rawat:
@helfi92 feel free to have a look once I get time 👍

